PR #433, opened 2020-11-17
Fix admin console in public room

gparant:


moufmouf (changes requested):
Hey @gparant ,

I see issues with the way you are dealing with this problem.
If you remove the JWT token, you will have this kind of issues:

If a private map points to a public map and the public map points to the private map, when going back and forth between the maps, you will face problems.

Let's imagine:

A user is in a private map
He walks to the public map
He goes back to the private map => he no longer has the credentials.

We should deal with this issue completely differently in my opinion.
The real problem is that we store tags in the JWT token. Tags are actually dependent on the user AND the map.
Tags are already correctly sent in the `RoomUserJoined`. What we need to do:

1- make sure we only send the correct tags (so fix the back)
2- make sure we don't store tags in local storage (if we do)

gparant:
Yes I think that it coul be a good solution to check JWT in back. But the problem still same, if user is in private room and go on public room, how we define the right of user ?

I think that this fix concern only event private map(s). I think that tommorrow, if we use private and public rooms, we must change the conduct to check authentification user when siwthed between private and public room ?

PR #447, opened 2020-11-22
Manage users when room is full

gparant:
This merges request permit to send message of full room when the user tries to connect to the room.

image

Yes, I don't close the connexion. I think that the user can navigate in the room could be a good thing. Don't worry, If the room is full, the message sent will be never take by the back end. 

You tell me that the user could join a JITSI room or open a web site. Yes, I think that it's not a problem, this is managed in the front end. So we still limit interaction and calculate in the back end.

### DIFF
--- a/back/src/Controller/IoSocketController.ts
+++ b/back/src/Controller/IoSocketController.ts
@@ -165,9 +165,6 @@ export class IoSocketController {
                         let memberTags: string[] = [];
                         let memberTextures: CharacterTexture[] = [];
                         const room = await socketManager.getOrCreateRoom(roomId);
-                        if(room.isFull){
-                            throw new Error('Room is full');
-                        }
                         if (ADMIN_API_URL) {
                             try {
                                 const userData = await adminApi.fetchMemberDataByUuid(userUuid);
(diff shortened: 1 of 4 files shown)


back/src/Controller/IoSocketController.ts
```diff
@@ -240,9 +237,26 @@ export class IoSocketController {
...
                 // Let's join the room
                 const client = this.initClient(ws); //todo: into the upgrade instead?
-                socketManager.handleJoinRoom(client);
 
                 //get data information and show messages
+                const room = socketManager.getRoomById(client.roomId);
+                if(room && room.isFull){
+                    socketManager.emitSendUserMessage({
+                        userUuid: client.userUuid,
+                        message: `<p style="text-align: center; font-size: 40px;">
```
Kharhamel, line 246: pourquoi envoyer le texte de l'erreur en html depuis le serveur? Pourquoi ne pas le mettre directement dans le client, et envoyer depuis le back un message plus court?

Kharhamel (changes requested):

Kharhamel:
Already done

PR #448, opened 2020-11-22
Report button discussion

gparant:
Disabled report button of discussion on anonymous connection.


gparant:
Public test on this deploy public map ✅ 
Private test on private map from admin server ✅ 

I merge this fix

PR #527, opened 2020-12-29
feat: adds property openWebsitePolicy property to set "allow" property in iframe

oliverlorenz:
Hello and greetings from rc3 ;)

First things first: I love your project! It is (almost) exactly what I always imagined and what I want to have. That's why I've spent quite some time playing with it.

One of my goals was to include an external jitsi via an iframe. That's not a problem in itself, but it brings the feature policy into it: https://developer.mozilla.org/en-US/docs/Web/HTTP/Feature_Policy.

Meaning: I can't share my microphone and camera within the iframe. I'm really no expert at this, but here's what should solve the problem:

I introduced a new property on the layer: Called `openWebsitePolicy`. This allows to set the "allow" attribute of the iframe. This allows the mapper to explicitly allow what the iframe is allowed to do. 
To include Jitsi via iframe, it should be sufficient to set the property `openWebsitePolicy` with `camera; microphone` in addition to the property `openWebsite` to enable camera and microphone.

Unfortunately I can't test this properly. Maybe you see another way to solve this. 

Or: Maybe you can review this and put it in one of the next releases.  That would make me very happy!

Many greetings
Oliver

### DIFF
--- a/front/src/Phaser/Game/GameScene.ts
+++ b/front/src/Phaser/Game/GameScene.ts
@@ -604,7 +604,7 @@ export class GameScene extends ResizableScene implements CenterListener {
                 coWebsiteManager.closeCoWebsite();
             }else{
                 const openWebsiteFunction = () => {
-                    coWebsiteManager.loadCoWebsite(newValue as string);
+                    coWebsiteManager.loadCoWebsite(newValue as string, allProps.get('openWebsitePolicy') as string | undefined);
                     layoutManager.removeActionButton('openWebsite', this.userInputManager);
                 };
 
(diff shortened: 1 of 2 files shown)


oliverlorenz:
Another person who implemented something similar: https://github.com/thecodingmachine/workadventure/pull/602

Kharhamel:
Thanks for your feedback! Sorry we didn't respond sooner, there is a lot of contributions and we are short on time. It looks like a privacy/security issue rather than simple code, so I will have to look into it before beeing sure it can be merged

albancho:
Exactly what we are waiting for, so we could use openWesbsite to a personnal HTML file, including gaming iframe (like https://skribbl.io/) and a jitsi together !

gparant:
Thank you! 💪

albancho:
This is live, and works very well ! Thanks !

PR #561, opened 2021-01-06
Verify URL in message

gparant:


Informatic:
This introduces an XSS vulnerability exploitable via: `https://"onmouseover="alert(1)`

PR #576, opened 2021-01-08
back,pusher,uploader: cleanup dependencies

Informatic:
Move typescript/ts-node-dev to dev dependencies, remove unused
dependencies.

PR #577, opened 2021-01-08
*: Dockerfiles cleanup, docker-compose.prod.yaml

Informatic:
New docker-compose.prod.yaml should provide a production-ish deployment of WorkAdventure for people willing to host it themselves.

All backend services are now split into 3 phases: protobuf build, typescript build, final production image (with only production dependencies present), hopefully giving smaller container images in the end. (#576 will surely help with that)

Biggest issue with this setup is that frontend URLs are customized in buildtime via args, instead of runtime. This could be brought back without bringing back whole javascript/nodejs environment into frontend container by extracting configuration into runtime variable set in index.html and then appending/modifying index.html on startup using envsubst or similar tool. I can follow up with that solution, if I get some initial approval.

### DIFF
--- a/docker-compose.prod.yaml
+++ b/docker-compose.prod.yaml
@@ -1,0 +1,62 @@
+version: "3"
+services:
+  reverse-proxy:
+    image: traefik:v2.3.7
+    command:
+      - --providers.docker
+      - --entryPoints.web.address=:80
+      - --entryPoints.websecure.address=:443
+    ports:
+      - "80:80"
+      - "443:443"
+    depends_on:
+      - back
+      - front
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+
+  front:
+    build:
+      context: .
+      dockerfile: front/Dockerfile
+      args:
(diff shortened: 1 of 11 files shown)


docker-compose.prod.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - front
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+
+  front:
+    build:
+      context: .
+      dockerfile: front/Dockerfile
+      args:
+        BASE_DOMAIN: ${BASE_DOMAIN:-workadventure.localhost}
```
evrardjp, line 23: Can you update the .env to add BASE_DOMAIN (defaulting to workadventure.localhost), and change the readme to mention it?

evrardjp (commented):
Looks like a good idea

sebkur:
@Informatic sorry to highjack the PR, I'm trying to use this in production as well and I guess your changes fix the issue I'm having where the main.*.js is about 23MB in size. Do you know which part of you PR fixes that?

PR #578, opened 2021-01-08
front: expose all services on the same domain name

Informatic:
This is a followup to #577, exposing both pusher and front on the same (default) domain vhost would greatly ease up deployment. This also makes front use `window.location.host` if environment API_URL is empty.

### DIFF
--- a/docker-compose.prod.yaml
+++ b/docker-compose.prod.yaml
@@ -1,0 +1,62 @@
+version: "3"
+services:
+  reverse-proxy:
+    image: traefik:v2.3.7
+    command:
+      - --providers.docker
+      - --entryPoints.web.address=:80
+      - --entryPoints.websecure.address=:443
+    ports:
+      - "80:80"
+      - "443:443"
+    depends_on:
+      - back
+      - front
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+
+  front:
+    build:
+      context: .
+      dockerfile: front/Dockerfile
+      args:
+        BASE_DOMAIN: ${BASE_DOMAIN:-workadventure.localhost}
+    labels:
+      - "traefik.http.routers.front.rule=Host(`play.${BASE_DOMAIN:-workadventure.localhost}`)"
+      - "traefik.http.routers.front.entryPoints=web"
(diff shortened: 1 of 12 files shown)


docker-compose.prod.yaml
```diff
@@ -0,0 +1,65 @@
+version: "3"
+services:
+  reverse-proxy:
+    image: traefik:v2.0
```
TuxCoder, line 4: Why you not use a newer image, this is a year old
https://hub.docker.com/layers/traefik/library/traefik/v2.0/images/sha256-384830d8a7692d3169768e284b49f7a96a73e74d8d3a9f4b143d9eac86f3b919?context=explore

front/Dockerfile
```diff
@@ -1,15 +1,28 @@
...
+ARG BASE_DOMAIN=workadventure.localhost
+ARG API_URL=pusher.$BASE_DOMAIN
+ARG UPLOADER_URL=uploader.$BASE_DOMAIN
+ARG ADMIN_URL=admin.$BASE_DOMAIN
+
+RUN API_URL=$API_URL UPLOADER_URL=$UPLOADER_URL ADMIN_URL=$ADMIN_URL \
+    yarn run build
+
+# final production image
+FROM nginx:1.18.0-alpine@sha256:da3716611fb965f3fda1f3281882baeb2760ca8bb7317f1d22ed45e75570827b
```
TuxCoder, line 26: same here, looks like this is still supported (last build 25days agoi) but there is a newer version,
nginx:1.19-alpine

TuxCoder (commented):
not sure if you wanna fix it in the original PR, or in a follow up

front/Dockerfile
Informatic: Good point, copied it over off some of my older production images.

docker-compose.prod.yaml
Informatic: Done.

pizkaz:
Hmm if I understand correctly, this PR is obsolete, because single domain deployment can be done another way now, right?

Kharhamel:
Yes it is, @moufmouf created a docker-compose-sing-domain.yml. Might as well close this PR. Thanks for your help

PR #597, opened 2021-01-14
add remote files and streams to playAudio

psy:
Add the possibility to include remote files and streams to the playAudio feature.


Kharhamel:
Thanks for the PR!

Issue #465, opened 2020-11-27
Integration of BigBlueButton


datenreisen:
Thank you very much for this great project.

I would like to suggest to integrate the video conferencing platform BigBlueButton as Jitsi Meet. Jitsi Meet is very easy to use for smaller meetings. BigBlueButton has advantages for larger events, seminars etc..

The API calls are described [here](https://docs.bigbluebutton.org/dev/api.html). For the "create" API call at least custom properties for BBBmeetingID, BBBattendeePW and BBBmoderatorPW are needed. This "create" API call follows a "join" API call together with the fullName and the desired password (selectable joining as moderator or attendee). Best would be, all create parameters could be optional provided as custom properties (BBB[nameofparameter]).

In addition a secret and a baseURL has to be defined on the server itself, which are needed for the checksum calculation and the API call.

Thanks in advance!

lunar-debian:
The code in #602 is rough but kinda working. Feel free to have a look and advice on how to improve the code.

PR #599, opened 2021-01-14
WIP: Integrate with BigBlueButton

znerol:
Addresses: #465 

- [x] Implement property change handler for new `bbbRoom` property in `GameScene`.
- [x] Add necessary messages in protocol in `messages.proto`.
- [x] Add stub handlers in both backends (`pusher` and `back`). Returns stub join url to client.
- [x] Add stub frontend in `BbbFactory.ts`. Logs stub join url received from backend to console.
- [ ] Proper backend implementation which creates a meeting via BBB API if necessary and generates the join url.
- [ ] Proper frontend which either loads the join url in an `iframe` or displays a button to open BBB in a new tab.

### DIFF
--- a/back/src/RoomManager.ts
+++ b/back/src/RoomManager.ts
@@ -9,6 +9,7 @@ import {
     PlayGlobalMessage,
     PusherToBackMessage,
     QueryJitsiJwtMessage,
+    QueryBbbJoinMessage,
     ReportPlayerMessage,
     RoomJoinedMessage,
     ServerToAdminClientMessage,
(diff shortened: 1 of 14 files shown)


lunar-debian:
I have been working on this too. I'll try to publish a branch today.

znerol:
Oh cool! Let's join forces then. I'm still a bit unsure on how to inject the room configuration into `pusher` and `back`.

lunar-debian:
See #602. It’s not working yet. I'm just publishing early to avoid duplicating our efforts. :)

znerol:
Closing in favor of #602.

PR #602, opened 2021-01-15
WIP: Add support for BigBlueButton

### LINKED ISSUE #465: Integration of BigBlueButton

lunar-debian:
This is still a work in progress. If something breaks, you are free to keep the pieces. Co-authored with @znerol. 

Fixes #465.

What this does:

* Add `BBB_URL` and `BBB_SECRET` as environment variable to configure the backend.
* Add support for the `bbbMeeting` string property to tiles. This will start a meeting with the specified identifier.
* Add support for extra string properties starting with `userdata-bbb_` to [configure the behavior of BBB HTML5 client](https://docs.bigbluebutton.org/2.2/customize.html#passing-custom-parameters-to-the-client-on-join). This should allow to map a theater with a stage and an audience having different settings for camera/microphone.

Left to be done:

* [x] Build a map to test the "theater" settings mentioned above.
* [ ] Figure out how to make admins join the meeting as moderators.
* [ ] Have the client send microphone and camera state when joining to the server so we can join with matching settings.
* [ ] Clean up. Some of the behavior is common with Jitsi and should probably be factored out. The code might not be at the right place server-side as well.

### DIFF
--- a/front/src/Phaser/Game/GameScene.ts
+++ b/front/src/Phaser/Game/GameScene.ts
@@ -518,6 +518,13 @@ export class GameScene extends ResizableScene implements CenterListener {
                 this.startJitsi(room, jwt);
             });
 
+            /**
+             * Triggered when we receive the URL to join a meeting on BBB
+             */
+            this.connection.onBBBMeetingClientURL((meetingId: string, clientURL: string) => {
+                this.startBBBMeeting(meetingId, clientURL);
+            });
+
             // When connection is performed, let's connect SimplePeer
             this.simplePeer = new SimplePeer(this.connection, !this.room.isPublic, this.playerName);
             this.GlobalMessageManager = new GlobalMessageManager(this.connection);
@@ -635,6 +642,14 @@ export class GameScene extends ResizableScene implements CenterListener {
                 }
             }
         });
+	this.gameMap.onPropertyChange('bbbMeeting', (newValue, oldValue, allProps) => {
+            if (newValue === undefined) {
+                layoutManager.removeActionButton('bbbMeeting', this.userInputManager);
+                this.stopBBBMeeting();
+            } else {
+                this.connection.emitJoinBBBMeeting(newValue as string, allProps);
+            }
+	});
         this.gameMap.onPropertyChange('jitsiRoom', (newValue, oldValue, allProps) => {
             if (newValue === undefined) {
                 layoutManager.removeActionButton('jitsiRoom', this.userInputManager);
@@ -1219,5 +1234,19 @@ export class GameScene extends ResizableScene implements CenterListener {
         mediaManager.removeTriggerCloseJitsiFrameButton('close-jisi');
     }
 
+    public startBBBMeeting(meetingId: string, clientURL: string): void {
+        this.connection.setSilent(true);
+        mediaManager.hideGameOverlay();
+        const clientOrigin = (new URL(clientURL)).origin;
+        coWebsiteManager.loadCoWebsite(clientURL, iframe => {
+            iframe.allow = `microphone ${clientOrigin}; camera ${clientOrigin}`;
+        });
+    }
+
+    public stopBBBMeeting(): void {
+        coWebsiteManager.closeCoWebsite();
+        mediaManager.showGameOverlay();
+        this.connection.setSilent(false);
+    }
 
 }
--- a/back/package.json
+++ b/back/package.json
@@ -38,6 +38,7 @@
   "homepage": "https://github.com/thecodingmachine/workadventure#readme",
   "dependencies": {
     "axios": "^0.21.1",
+    "bigbluebutton-js": "^0.1.1",
     "body-parser": "^1.19.0",
     "busboy": "^0.3.1",
     "circular-json": "^0.5.9",
--- a/back/src/Enum/EnvironmentVariable.ts
+++ b/back/src/Enum/EnvironmentVariable.ts
@@ -10,6 +10,8 @@ const CPU_OVERHEAT_THRESHOLD = Number(process.env.CPU_OVERHEAT_THRESHOLD) || 80;
 const JITSI_URL : string|undefined = (process.env.JITSI_URL === '') ? undefined : process.env.JITSI_URL;
 const JITSI_ISS = process.env.JITSI_ISS || '';
 const SECRET_JITSI_KEY = process.env.SECRET_JITSI_KEY || '';
+const BBB_URL: string|undefined = (process.env.BBB_URL === '') ? undefined : process.env.BBB_URL;
+const BBB_SECRET = process.env.BBB_SECRET || '';
 const HTTP_PORT = parseInt(process.env.HTTP_PORT || '8080') || 8080;
 const GRPC_PORT = parseInt(process.env.GRPC_PORT || '50051') || 50051;
 export const SOCKET_IDLE_TIMER = parseInt(process.env.SOCKET_IDLE_TIMER as string) || 30; // maximum time (in second) without activity before a socket is closed
@@ -28,5 +30,7 @@ export {
     CPU_OVERHEAT_THRESHOLD,
     JITSI_URL,
     JITSI_ISS,
-    SECRET_JITSI_KEY
+    SECRET_JITSI_KEY,
+    BBB_URL,
+    BBB_SECRET
 }
--- a/back/src/RoomManager.ts
+++ b/back/src/RoomManager.ts
@@ -9,6 +9,7 @@ import {
     PlayGlobalMessage,
     PusherToBackMessage,
     QueryJitsiJwtMessage,
+    JoinBBBMeetingMessage,
     ReportPlayerMessage,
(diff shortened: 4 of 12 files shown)


znerol:
> Current issues:
> * This worked only after changing allowRequestsWithoutSession=true in bigbluebutton.properties. Probably related to the JSESSIONID cookie.
> * BBB is unable to get the permission to use the microphone.

We might want to simply display a link/button *Join meeting* which opens BBB in a new tab instead of trying to load it inside an `iframe` . Browser vendors are aggressively fighting clickjacking and fingerprinting practices and as a result framing content will be less and less an option. Which is good for privacy anyway.

znerol:
Let's continue with this PR. All the necessary bits are in place already.

lunar-debian:
Screenshot in action:
image

front/src/Phaser/Game/GameScene.ts
```diff
@@ -1219,5 +1234,18 @@ export class GameScene extends ResizableScene implements CenterListener {
         mediaManager.removeTriggerCloseJitsiFrameButton('close-jisi');
     }
 
+    public startBBBMeeting(meetingId: string, clientURL: string): void {
+        this.connection.setSilent(true);
+        mediaManager.hideGameOverlay();
+        coWebsiteManager.loadCoWebsite(clientURL, iframe => {
+            iframe.allow = `microphone ${clientURL}; camera ${clientURL}`;
+        });
```
znerol, line 1242: Nice!

lunar-debian: Actually, I don’t think this is right and it’s probably why it’s not working on Firefox. It should be of the form `https://bbb.example.org/` and not `https://bbb.example.org/bigbluebutton/…` I believe `clientURL` needs to be stripped away of its path.

znerol:
Potential fix for Firefox for the camera / microphone permissions inside the `iframe`:

```typescript
        const clientOrigin = (new URL(clientURL)).origin;
        coWebsiteManager.loadCoWebsite(clientURL, iframe => {
            iframe.allow = `microphone ${clientOrigin}; camera ${clientOrigin}`;
        });
```

I did some research on how other integrations are battling the `JSESSIONID` issue. [BBB app for nextcloud](https://github.com/sualko/cloud_bbb/blob/ef413ac79329ec73c24aa06a72e5a6c0a5e45eec/lib/BigBlueButton/API.php#L81-L98) uses the `redirect` method explicitly. I.e., it sends the `api/join` url to the client and that will set the cookie and redirect to the HTML5 client. Same story in [greenlight](https://github.com/bigbluebutton/greenlight/blob/752b192ef9360fd129ffc988b3cd2011d42ed9e2/app/controllers/concerns/bbb_server.rb#L45-L59) uses redirect implicitly (redirect defaults to `true`). Maybe we should do the same, that way there is no need to tweak `allowRequestsWithoutSession`.

Regarding embedding the client inside an `iframe`. I fear that this will lead to more issues down the road even if `allowRequestsWithoutSession` is set to `true`. Especially for content which is then framed within the bbb application (e.g., etherpad). So I guess I might just work on an alternative way to open bbb in a new window/tab. Ideally such things should be configurable via map properties anyway.

znerol:
Pushed a couple of fixes into my fork of this branch: https://github.com/znerol-forks/workadventure/tree/bbb-integration

This works nicely with Firefox now. `userdata-bbb_XYZ` custom properties can be set on the map and are passed to the URL generator. Please merge/cherry-pick whatever you think is useful from over there.

lunar-debian:
> I did some research on how other integrations are battling the `JSESSIONID` issue. [BBB app for nextcloud](https://github.com/sualko/cloud_bbb/blob/ef413ac79329ec73c24aa06a72e5a6c0a5e45eec/lib/BigBlueButton/API.php#L81-L98) uses the `redirect` method explicitly. I.e., it sends the `api/join` url to the client and that will set the cookie and redirect to the HTML5 client. Same story in [greenlight](https://github.com/bigbluebutton/greenlight/blob/752b192ef9360fd129ffc988b3cd2011d42ed9e2/app/controllers/concerns/bbb_server.rb#L45-L59) uses redirect implicitly (redirect defaults to `true`). Maybe we should do the same, that way there is no need to tweak `allowRequestsWithoutSession`.

One downside of doing this which I just stumbled upon while testing your latest branch is that it requires `BBB_URL` to be set to a publicly reachable endpoint in WA. But I guess this is how it should be done anyway.

lunar-debian:
> Pushed a couple of fixes into my fork of this branch: https://github.com/znerol-forks/workadventure/tree/bbb-integration

Thanks a lot! :) I've squashed your changes in my commit. The only change I did was to use the spread operator instead of `Object.assign`.

lunar-debian:
I’ve done a test map with a stage and an audience. It works! :)

One thing though: I believe people on the stage should have moderator privileges. Otherwise, they can’t use presenter features. Also, the audience needs to login with `attendeePW`, otherwise, they can join with their microphone.

So, in order to support this use-case, we probably need to handle another property and switch which password is used to login accordingly.

What slightly annoys me is that jumping on a stage is enough to kick other participants out, but I guess that’s life. Hopefully we’ll find a way to make BBB reject users once they have been banned from WorkAdventure.

Relevant layer properties : [stage.json](https://github.com/thecodingmachine/workadventure/files/5832391/stage.json.txt), [room.json](https://github.com/thecodingmachine/workadventure/files/5832392/room.json.txt)

lunar-debian:
So… after some real life testing this evening, it feels that there definitely is some limits to how BBB is a good match for WorkAdventure. The main issue is that joining a BBB meeting is slow. It can take up to a minute on a moderately loaded server to have working sound, video, microphone and camera.

I’m starting to wonder if [Galene](https://github.com/jech/galene) would not be a much better for WorkAdventure…

jech:
You'll find a summary description of how to write a new frontend for Galène here: https://galene.org/README.FRONTEND.html
The protocol is described here: https://galene.org/README.PROTOCOL.html

Kharhamel:
First, thank you for the PR. Sorry we didn't respond sooner, we have some issues reading all the contributions. I didn't have time to test your branch, do you think your feature is functionnal or not?
You said Galene might be a better match for wa, could you explain why?

znerol:
In my opinion adding support for BigBlueButton to workadventure would be useful for organizations which are already using BBB. It certainly is not useful as a replacement for Jitsi. Despite the claims of the BBB maintainers, setting up and maintaining the application properly (and securely) is time consuming and hard work. With enough experience it is possible to provide a good and responsive service (where it only takes seconds to setup mic and video).

The PR is ready for testing it is not quite ready for merging. I'd appreciate advice from the maintainers on how to differentiate between moderators and viewers. From the comments in the source I guess this problem isn't properly solved in the jitsi code path neither. So one way would be to tackle that problem in a different issue/PR before and then use that mechanism here as well.

lunar-debian:
> You said Galene might be a better match for wa, could you explain why?

WorkAdventure currently does not integrate with a SFU. While BBB is nice, the web client is quite heavy and getting inside a room takes time. This does not make a smooth experience. The situation is slightly better with Jitsi but not that much. Galene client is really lightway, so it might be interesting to see how much better it would be.

In the end, what I really want is proper integration with a SFU so that being in a group conversation does not rely on an iframe. Given current WorkAdventure architecture, this would require significant changes.

znerol:
opened #654

fcecagno:
Hi guys, we're interested to complete this work so it can get into the core. We're one of the companies involved in the BigBlueButton ecosystem, and we'd love to have a really nice integration between BigBlueButton and Workadventure - and we're willing to do the work to complete it.
We've been using BigBlueButton on iframe for a long time, so IMO there's no challenge there. Also, we can connect users with audio and video within seconds, just like Jitsi does today, so I'm pretty sure the experience can be similar if we want to.

From the project's point of view, what would you like to have in this integration in order to merge it into the core?

fcecagno:
Hi everyone, sharing with you some preliminary results. You can see how fast it loads the room with audio and video, pretty cool. We're planning to open a pull request in the upcoming days. Any considerations or suggestions are welcome.

https://user-images.githubusercontent.com/502164/155827505-28743622-1093-4b4c-98e4-5727c949623b.mp4

nolway:
In development in https://github.com/thecodingmachine/workadventure/pull/1948

PR #616, opened 2021-01-18
Translate English to French

lunar-debian:
While this does not implement proper localization, this is an attempt to get the front-end in French as much as currently possible, at least to identify limitations and which strings should be internationalized.

One major issue for now:

* The “avatar” bitmap font only contains the basic ASCII set, meaning one has to use “entree” and not the properly accented “entrée”.

### DIFF
--- a/front/dist/resources/html/gameMenu.html
+++ b/front/dist/resources/html/gameMenu.html
@@ -21,22 +21,22 @@
 
     <main>
         <section>
-            <button id="shareButton">Share url</button>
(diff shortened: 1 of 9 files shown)


Kharhamel:
We definitly are not going to make french the default language since it  is used outside of France, so I cannot merge this, sorry.

I don't think we plan on doing a proper localization for now, unless we find a way to do it painlessly on phaser.

gparant:
I close this merge request. We still available if you have any question.

Thanks 💪

PR #658, opened 2021-01-25
Lock user banned

gparant:
Use error scene and show banned message

### DIFF
--- a/front/src/Phaser/Game/GameScene.ts
+++ b/front/src/Phaser/Game/GameScene.ts
@@ -1237,14 +1237,13 @@ export class GameScene extends ResizableScene implements CenterListener {
     }
 
     private bannedUser(){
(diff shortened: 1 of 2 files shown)


front/src/Phaser/Game/GameScene.ts
```diff
@@ -1235,5 +1236,15 @@ export class GameScene extends ResizableScene implements CenterListener {
         mediaManager.removeTriggerCloseJitsiFrameButton('close-jisi');
     }
 
+    private bannedUser(){
+        this.scene.start(ErrorSceneName, {
```
Kharhamel, line 1240: you should call cleanupClosingScene() before changing to errorScene. this means that you won't need the methods stopJitsi and closeCoWebsite after it.

Kharhamel (changes requested):

PR #669, opened 2021-01-27
store "muted" state of audio player in local store

tabascoeye:
The state of "muted" of the AudioPlayer should be stored in localStore just as the volume level is storeg already.

(I also renamed the key for the volume so that is is more unique)


Kharhamel (changes requested):
Please put every reference to localStorage in the dedicated class LocalUserStore.

tabascoeye:
I copied from the way the "volume" is already stored in localUserStore. You're saying that is already wrong?

Kharhamel:
Yes. There is some shaddy code in wa like in all projects. Some people don't mind having localStorage (or other important global objects) accesses hidden in all the project, even though it is a big code smell.

tabascoeye:
whops. Something went wrong merging. Closing this and will reopen a new PR. Sorry about that

PR #685, opened 2021-01-29
save "mute" state in localStore

tabascoeye:
When the "background" audio player is being muted, this state was not save in LocalStore of the browser.
This PR add that and refactors the use of localStore so that also the volume is stored and received via getter/setter functions from the central class


gparant:
It's very good idea, thank you 💪

pizkaz:
An updated version of this is contained in #825.

tabascoeye:
please merge #825 from @pizkaz. I will close this

PR #688, opened 2021-01-30
disable automatic gain control on microphone

tabascoeye:
AGC really makes the soundqulity way worse than it could be.
Especially since it is so noticeable when entering a Jitsi where AGC is disabled (and suddenly some people sound much better).
Also noise suppression might not be a good idea (but I may be wrong on this one)

In the long run, this should probably be configurable. On the other hand the setting changes the audio _going out_ from a user, so it might be hard for him to judge which settings are a good or a bad idea…

### DIFF
--- a/front/src/WebRtc/MediaManager.ts
+++ b/front/src/WebRtc/MediaManager.ts
@@ -18,7 +18,7 @@ let videoConstraint: boolean|MediaTrackConstraints = {
     resizeMode: 'crop-and-scale',
     aspectRatio: 1.777777778
 };
-let audioConstraint: boolean|MediaTrackConstraints = {
+const audioConstraint: boolean|MediaTrackConstraints = {
     //TODO: make these values configurable in the game settings menu and store them in localstorage
     autoGainControl: false,
     echoCancellation: true,
@@ -243,7 +243,7 @@ export class MediaManager {
     }
 
     public enableMicrophone() {
-        this.constraintsMedia.audio = true;
+        this.constraintsMedia.audio = audioConstraint;
 
         this.getCamera().then((stream) => {
             //TODO show error message tooltip upper of camera button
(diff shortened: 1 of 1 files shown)


Kharhamel:
Good idea! Do you want to try to create a audio settings menu in this PR or in another?

front/src/WebRtc/MediaManager.ts
```diff
@@ -18,6 +18,12 @@ let videoConstraint: boolean|MediaTrackConstraints = {
     resizeMode: 'crop-and-scale',
     aspectRatio: 1.777777778
 };
+const audioConstraint: boolean|MediaTrackConstraints = {
```
Kharhamel, line 21: Can't you just type it as MediaTrackConstraints?bool should not be necessary anymore

Kharhamel (changes requested):

front/src/WebRtc/MediaManager.ts
tabascoeye: I don't know TypeScript well enough

Kharhamel:
Sorry for leaving this PR for so long, I completely forgot about it. The typing is weird but it will fix it later. Thanks a lot!

front/src/WebRtc/MediaManager.ts
```diff
@@ -18,6 +18,12 @@ let videoConstraint: boolean|MediaTrackConstraints = {
...
+    //TODO: make these values configurable in the game settings menu and store them in localstorage
+    autoGainControl: false,
+    echoCancellation: true,
+    noiseSuppression: false
```
tabascoeye, line 25: this turned out to be a really bad idea. Should be re-enabled!

tabascoeye: I mean noiseSuppression. autoGainControl should still stay false!

PR #690, opened 2021-01-30
Update Loading page

gparant:
Add WorkAdventure LOGO


moufmouf:
This seems to be failing:
image

gparant:
> This seems to be failing:
> image

I think that I will verify if the logo is loaded and is not, show loading text.

moufmouf:
Top! :+1:

moufmouf:
Note: I think we could do even better:

- [ ] Better centering of the logo and loading block vertically
- [ ] The loading progress bar should not be that thick and should not take the whole width of the screen (maybe 1/3 of the screen and 32px height?)

moufmouf (approved):

PR #703, opened 2021-02-03
Social network link anonymous connexion

gparant:
In menu, add social newtork link when user is connected in anonymous mode


front/src/Phaser/Menu/MenuScene.ts
```diff
@@ -115,6 +117,10 @@ export class MenuScene extends Phaser.Scene {
             const adminSection = this.menuElement.getChildByID('adminConsoleSection') as HTMLElement;
             adminSection.hidden = false;
         }
+        if (connectionManager.getConnexionType === GameConnexionTypes.anonymous){
```
gparant, line 120: Add promise to check if connection is good

PR #717, opened 2021-02-09
Add STUN_SERVER env variable with fallback

swiesend:
default: "stun:stun.l.google.com:19302"

Make STUN_SERVER available for app config

Change also the STUN_SERVER for ScreenSharingPeer


moufmouf (approved):
Thanks a lot for this PR.
Looks good to me!
:+1:

PR #736, opened 2021-02-13
Fix offsets of SpeechBubble.ts

MisterErwin:
Since some recent optimizations, they are being added on top of the player, which makes the offset redundant.

(Also they appear to be unused - but I wanted to play around with them.)

### DIFF
--- a/front/src/Phaser/Entity/SpeechBubble.ts
+++ b/front/src/Phaser/Entity/SpeechBubble.ts
@@ -14,7 +14,7 @@ export class SpeechBubble {
         const bubbleWidth = bubblePadding * 2 + text.length * 10;
         const arrowHeight = bubbleHeight / 4;
 
-        this.bubble = scene.add.graphics({ x: player.x + 16, y: player.y - 80 });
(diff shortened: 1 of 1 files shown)


Kharhamel:
It is an old feature I used to do various tests, before we create video and text based chats. How would you like to use it?

MisterErwin:
Almost thought so - I played around a bit with your (awesome) project and tried (among others) to add non-verbal reactions.
(To be fair, I used the speech bubbles initially to test the menu in the first screen - but I figured why not fix the bubbles so they are visible again if someone else also attempts to use them)

image
image

moufmouf:
Thanks a lot for this PR.

Hey! This feature is excellent?
Is there anywhere we can have a look at it?

MisterErwin:
No problem - glad I can help out.

> Is there anywhere we can have a look at it?

I'll make it a PR (most likely a draft one) once I have a better insight into the client/server communication - as it is currently just clientside.

moufmouf:
Client / server communication is not completely trivial. Let me know if you need some help figuring it out!

PR #777, opened 2021-02-24
Fix admin message and ban functionality

s00500:
When trying out the ban and sendAdminMessage functions I got an error saying that GRPC expected a ServerToClientMessage. This happened on both requests. I changed from the SubToPusherMessage to a ServerToClientMessage and it seems to work nicely, so I suspect this was a bug :-)


pizkaz:
I just tried to rebase this onto current develop: It seems this was fixed.

s00500:
Yep, I think so too, I have seem a few changes, have not checked in more detail yet though, anyway this can savely be closed for now

PR #780, opened 2021-02-27
Make name entry on mobile browsers possible

vmx:
Prior to this change, there wasn't a way in mobile browsers to show
the virtual keyboard in order to enter the player name.

With this change, the name entry is now done via a hidden input field,
which enables the use if a virtual keyboard.

Mobile browsers are not special cased, the desktop browsers will also
use the input field.


vmx:
I just found out that this scene doesn't work properly when it is accessed from within the game through `Edit name`. Somehow the `pointerdown` event isn't triggered.

Kharhamel:
@gparant is already redoing the loginScene from scratch design wise. Still, thanks for contributing!

PR #783, opened 2021-03-01
Improve visibility of close button for embedded web pages

Moini:
The button (a white x) wasn't visible on white background. Now it is. Also is a little bit larger and changes color on hover.


Kharhamel:
Nice thank you!

PR #803, opened 2021-03-09
Mobile/touch support

pizkaz:
This combined work by @Palid, @tabascoeye and @pizkaz integrates a software joystick and adds touch support to login scenes.
With this patch set it is possible to use Workadventure on touch devices. \o/

Contrary to what commit metadata says, @Palid did most of the work for rC3. @TabascoEye and myself maintained the patches, tweaked stuff and implemented touch support for Jitsi/openWebsite triggers and movement speed following joystick "force", which is good for small displays because the normal walking speed is already too fast.

### DIFF
--- a/front/dist/resources/html/gameMenu.html
+++ b/front/dist/resources/html/gameMenu.html
@@ -41,6 +41,9 @@
             <button id="editGameSettingsButton">Settings</button>
         </section>
         <section>
+            <button id="showJoystick">Show joystick</button>
+        </section>
+        <section>
             <button id="sparkButton">Create map</button>
         </section>
         <section id="adminConsoleSection" hidden>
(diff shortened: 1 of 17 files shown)


pizkaz:
Hmm I have no idea what yarn is trying to tell me here. :-(
I back-ported these patches from our instance, which is running on Phaser 3.52.0. Could that be a problem?

tabascoeye:
It wants ```"esModuleInterop": true,"``` in ```front/tsconfig.json```

Kharhamel:
I think we will work on mobile support in our next sprint. Right now we are really focused on the payment features, so sorry if we don't yet give a lot of attention to this PR.

pizkaz:
Superseded by #873

Issue #801, opened 2021-03-09
Volume sound / music property ?


Jacques-Olivier-Farcy:
Hi,

As i am using some ambient sounds in my maps, will it be possible to set the volume of a sound with a property ?

pizkaz:
It is possible now. I believe this issue can be closed.

Jacques-Olivier-Farcy:
Hi @pizkaz ,

Do you mean the volume control UI that is opened when playing a sound in a zone ?

I was more asking to have the possibility in property to play the sound with a specific default volume

pizkaz:
Code: https://github.com/thecodingmachine/workadventure/commit/fdbcd98a9acc7b89e225c8294a881b00b4892d23
Docs: https://workadventu.re/map-building/special-zones (bottom of the page)

You're welcome. :-)

aresius:
Great feature but In the documentation I'm missing the property type and range of `audioVolume` (`float` or `int`, exact meaning of the number specified). Could you please add this information? Thanks.

pizkaz:
Unfortunately I can't because I do not have access to that documentation.

`audioVolume` is a float from 0 to 1.0

Moini:
Does this work for anyone? It doesn't work for me. Volume is set to 0.3 as default value for the slider, but the slider still starts at maximum value.

![2021-09-09_00-47](https://user-images.githubusercontent.com/3240233/132595747-74a89ac7-e52d-4694-bc01-da3e01b47bb8.png)

Moini:
Ah, I see... it works, but the slider isn't adjusted when entering the area. 

Also happens when I have turned down the volume entirely in one area, and then go into another area with sound. The slider is at 100%, even if no sound is played.

GRL78:
The audioVolume property exist for this purpose. Documentation here : https://staging.workadventu.re/map-building/special-zones.md#playing-sounds-or-background-music

Moini:
Yep, the bug is with the slider not updating, not the volume setting per se.

PR #808, opened 2021-03-12
Audio player improvements

pizkaz:
* adds two new map layer properties
  * `audioLoop` (boolean) &rArr; to loop audio files
  * `audioVolume` (number) &rArr; to play audio at lower volumes (implements #801)
* fixes automatic volume reduction in conversations
* fixes display of automatic volume reduction
* recycles #685
* small UI improvements

If the `audioVolume` property is set, the audio player uses either the value of the property or the last volume set by the user - whichever is smaller.


pizkaz:
Superseded by #825

PR #816, opened 2021-03-15
FEATURE: complete control over the iframe size

Kharhamel:
The iframes now have a small left margin. Holding click on it also to resize the iframe as you see fit

TODO:

* [x] add a neat visual in the iframe margin to convey the resize feature
* [x] How do we resize the game in  the background?


moufmouf (changes requested):
Pretty cool, but there is one issue:

It is breaking the responsiveness. On vertical displays, the iframe is broken and cannot be resized.

Kharhamel:
@moufmouf this is fixed now.

moufmouf (approved):

fbartels:
There seems to be another issue with this feature. When opening the iframe now Jitsi only uses half of the available size of the iframe. Only when removing https://github.com/thecodingmachine/workadventure/blob/1f85e548ce137a2db5a33588b72ae559ea390643/front/dist/resources/style/cowebsite.scss#L115 in the debugger jitsi will fill the full iframe.

image

PR #827, opened 2021-03-22
Fix Jitsi integration

pizkaz:
Jitsi iframes are rendered directly in the cowebsite div next to the
cowebsite > main element. This patch fixes the situation.


Kharhamel:
damn I forgot about that method. Thanks for the fix

PR #831, opened 2021-03-23
Missing commit in develop

pizkaz:
Looks as if this commit was done directly on master. I'd love to have that in develop, too. :-)


Kharhamel:
I merged master into develop, I think it is the better way to update.

PR #840, opened 2021-03-24
Replaced the hard-coded URLs by variables

ValdoTR:
Replace all the hard-coded URLs from the docker-compose file with env variables.
Add also one doc instruction about creating the env file needed by docker-compose.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -1,3 +1,12 @@
+# Domains
+DOMAIN=workadventure.localhost
+UPLOADER_SUBDOMAIN=pusher
+UPLOADER_SUBDOMAIN=uploader
+TURN_SUBDOMAIN=coturn
+FRONT_SUBDOMAIN=play
+MAPS_SUBDOMAIN=maps
+API_SUBDOMAIN=api
+
 DEBUG_MODE=false
 JITSI_URL=meet.jit.si
 # If your Jitsi environment has authentication set up, you MUST set JITSI_PRIVATE_MODE to "true" and you MUST pass a SECRET_JITSI_KEY to generate the JWT secret
--- a/.env.template
+++ b/.env.template
@@ -1,6 +1,6 @@
 # Domains
 DOMAIN=workadventure.localhost
-UPLOADER_SUBDOMAIN=pusher
+PUSHER_SUBDOMAIN=pusher
 UPLOADER_SUBDOMAIN=uploader
 TURN_SUBDOMAIN=coturn
 FRONT_SUBDOMAIN=play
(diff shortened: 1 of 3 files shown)


.env.template
```diff
@@ -1,3 +1,12 @@
+# Domains
+DOMAIN=workadventure.localhost
+UPLOADER_SUBDOMAIN=pusher
```
tabascoeye, line 3: should be PUSHER_SUBDOMAIN

docker-compose.yaml
```diff
@@ -26,13 +26,13 @@ services:
       JITSI_PRIVATE_MODE: "$JITSI_PRIVATE_MODE"
       HOST: "0.0.0.0"
       NODE_ENV: development
-      API_URL: pusher.workadventure.localhost
-      UPLOADER_URL: uploader.workadventure.localhost
-      ADMIN_URL: workadventure.localhost
+      API_URL: $UPLOADER_SUBDOMAIN.$DOMAIN
```
tabascoeye, line 29: should be $PUSHER_SUBDOMAIN.$DOMAIN

docker-compose.yaml
```diff
@@ -67,10 +67,10 @@ services:
     volumes:
       - ./pusher:/usr/src/app
     labels:
-      - "traefik.http.routers.pusher.rule=Host(`pusher.workadventure.localhost`)"
+      - "traefik.http.routers.pusher.rule=Host(`$UPLOADER_SUBDOMAIN.$DOMAIN`)"
```
tabascoeye, line 70: should be PUSHER_SUBDOMAIN

docker-compose.yaml
```diff
@@ -67,10 +67,10 @@ services:
...
       - "traefik.http.routers.pusher.entryPoints=web"
       - "traefik.http.services.pusher.loadbalancer.server.port=8080"
-      - "traefik.http.routers.pusher-ssl.rule=Host(`pusher.workadventure.localhost`)"
+      - "traefik.http.routers.pusher-ssl.rule=Host(`$UPLOADER_SUBDOMAIN.$DOMAIN`)"
```
tabascoeye, line 73: should be PUSHER_SUBDOMAIN

tabascoeye (changes requested):
looks nice, but UPLOADER should be PUSHER in a few places

ValdoTR:
Thank you @tabascoeye ! just a copy/past issue ;)
Will fix this.

ValdoTR:
I'm canceling thit because after thinking it, I don't think it really helps to allow the user to configure domains and subdomains, it might even mislead him.

PR #853, opened 2021-03-28
front: prevent navigation on backspace-key in LoginScene

lbehm:
Backspace input in LoginScene triggered navigation, because event wasn't handled in TextInput.ts. This affected only users with the classic browser feature for _navigate to previous page on backspace key_.

The problem may only occurred after phaser update.


pizkaz:
#873 probably also solves this issue, as TextInput.ts is rewritten in 0f43592fab8a26dbf8824964a04ca06826f78b8c.

Once that one is merged, could you verify that? Or alternatively name a Browser that maps `history.back()` to backspace?

lbehm:
In Firefox it can be found in `about:config`, setting `browser.backspace_action=2`

pizkaz:
Setting `browser.backspace_action=0` did the trick. Thx!

Seems like this is indeed fixed with said commit (new ID due to rebase: c7dcaec).

Kharhamel:
I tested it on develop and it is indeed fixed. Thanks!

PR #861, opened 2021-03-30
Use selected video quality during webcam initialization

joberthel:
The video frame rate setting is currently only working directly after you change it in the settings menu. During initialization (eg. after refresh) it tries to get the frame rate from the local storage with the key "select-video-quality". But it looks like this is some administration related setting.

I changed the code, so it gets the frame rate setting via the LocalUserStore class, which happens to use 20 as a default. So we can get rid of some lines of  code.


gparant:
Thanks @joberthel , I think too that we can get rid this lines of code 💪

PR #863, opened 2021-03-31
FEATURE: added the possibility to play emotes

Kharhamel:
Pretty simple feature: the ability to play an short emoticon to express your feelings to nearby users, without having to engage in a conversation.

TODO:

* [x] Add an emote menu to the game to be able to select which emote plays
* [x] Emotes played should be visible by other users
* [ ] Add the ability for a moderator to restrict/add emotes
* [ ] Add the ability for a map creator to add new emotes to his map using @moufmouf scripting feature.
* [x] Make all the emotes textures lazy-loaded
* [x] Find a nice set of emoticons to use (https://pipoya.itch.io/free-popup-emotes-pack may do the trick but it doesn't have a 'clap' emote :( )


tabascoeye:
Ideas:
* reduce complexity to maybe 6 important emotes
* have a "ring" style menu to select emote around the character on click on the character (this would also work on mobile/touch)
* optionally map those to number keys 1-6 as well
* ability to customize those 6 slots via map or backend

tabascoeye:
quick-and-dirty clap emote draft:
taba-clapemote

Kharhamel:
@tabascoeye that would be really cool. I am horrible when in comes to creating assets.

For the ring idea, it could become crowded very quickly once we start adding more emotes. However we could restrict it to the few important emotes. But we need to decide which emotes they need to be.

- I think "clap" is the most important one. It is the one that was requested.
- "Thumbs up" and "Thumbs down" may be next. They allow to express yes and no.
- I am not really sure what else. Stuff like "happy" or "sad" are not really emote in my opinion, they re more statuses (which is also a requested feature, but i don't think theses two should be mixed)

saschafoerster:
- ✋"lifting hand"🤚 to announce someone wants to speak?

tabascoeye:
more concept brainstorm:

* have single button for emotes "wheel/ring" menu similar to Secret of Mana on SNES
(see here for a reimplementation: https://www.reddit.com/r/IndieGaming/comments/iyfmvn/secret_of_manastyle_ring_menu/)
* after selection, the emote stays until emote button is pressed again (so you can have long "status" style emotes)
* I would still also have numer keys 1-9 as quick emote keys
* hitting same number as already showing cancels the emote
* new suggested emotes (additions to the ones mentioned above): 
** "AFK/idle" for people who want to indicate they are on a different tab or not at their computer -- texts exist in Pipoya set
** "!" for something like "hey check this out"  -- exists in Pipoya set
** "?" for "what is this?" -- exists in Pipoya set
** "zZZ" or similar to indicate "tired" "going to bed" etc -- exists in Pipoya set
** some emote to indicate "not ready to talk" or "please don't approach me"?
** opposite to the above "please talk to me" or "help me"?

Kharhamel:
I am not sure when it comes to 'statuses'. 
It would be cool to have them but right now the emotes takes a lot of space and they overwrite the player name. They are designed to be very visible. 

A status would have to be much more subtle: something like some small 'Z' next to the name for example. 
I think it would be better to design another system for statuses, with maybe some overlap with emotes, like the selection methods

tabascoeye:
* Y, U and I are working, O and P do not.
* animations are really slow and cannot be aborted

jonnytest1:
> Y, U and I are working

sounds like a cheesy pickup line :3

front/src/Connexion/EmoteEventStream.ts
```diff
@@ -0,0 +1,19 @@
...
+    emoteName: string,
+}
+
+class EmoteEventStream {
+
+    private _stream:Subject<EmoteEvent> = new Subject();
+    public stream = this._stream.asObservable();
+
+
+    onMessage(userId: number, emoteName:string) {
```
moufmouf, line 14: The `onXXX` naming convention is typically used to register callbacks (to listen to something).

What you are doing here is "firing" an event.

So maybe `fireEvent` would be a better name?

front/src/Phaser/Game/EmoteManager.ts
```diff
@@ -0,0 +1,67 @@
...
+    'emote-hand': {name: 'emote-hand', img: 'resources/emotes/hand-emote.png'},
+    'emote-thanks': {name: 'emote-thanks', img: 'resources/emotes/thanks-emote.png'},
+    'emote-thumb-up': {name: 'emote-thumb-up', img: 'resources/emotes/thumb-up-emote.png'},
+    'emote-thumb-down': {name: 'emote-thumb-down', img: 'resources/emotes/thumb-down-emote.png'},
+};
+
+export class EmoteManager {
+    
+    constructor(private scene: GameScene) {
+        emoteEventStream.stream.subscribe((event) => {
```
moufmouf, line 25: I think there is a memory leak here.

You must think about unsubscribing from the stream. Otherwise, the callback will be called even after you leave the Game.

moufmouf (changes requested):
All good to me except for a memory leak that needs to be fixed.

PR #878, opened 2021-04-07
FIX: improved the validation from localstorage for username and characterLayers

### LINKED ISSUE #875: Hack into GHOST mode
It's actually possible to "hack" the localstorage to become invisible. By setting an empty name, and editing the character's layers, it may result in a ghost.

People would only be noticed by the circle and the sound but they might think it's a bug. Therefore someone could spy some jitsi or peer2peer conversations.

Steps to reproduce set empty values in `playerName` & `characterLayers` properties of the local storage : 
image

Kharhamel:
this should fix #875 

EDIT: this PR add the env variable MAX_USERNAME_LENGTH. If no value is specified, the value 8 is picked instead.

### DIFF
--- a/front/src/Connexion/LocalUser.ts
+++ b/front/src/Connexion/LocalUser.ts
@@ -1,8 +1,27 @@
+import {MAX_USERNAME_LENGTH} from "../Enum/EnvironmentVariable";
+
 export interface CharacterTexture {
     id: number,
     level: number,
     url: string,
     rights: string
+}
+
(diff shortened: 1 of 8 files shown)


tabascoeye:
while you're introducing `maxUserNameLength`: would you mind introducing an env var for that so that we can easily customize that on our self-hosted instances? I usually set the allowed nickname length to 12 and would love to have that without changing source code

PR #902, opened 2021-04-17
add trigger action after closing co-website or jitsi

Lurkars:
Also not ideal because doubles code on some places, but I think in general it is a good improvement to show co-website or jitsi trigger after (manual) closing it. Currently, if you close website or jitsi and want to reopen it, you have to walk out of related tiles and walk in again. With this, you just can trigger a re-open.


jonnytest1:
should probably be toggleable (also we should aadd a callback to the scitp api if the website is closed)

Lurkars:
this is now deprecated with new Co-Website manager. Since I still want this feature for myself, I may open a new PR when I finished developing on this.

PR #917, opened 2021-04-20
Map village with new WA Office and features presented

ClementVieilly75:


jonnytest1:
🤔  which on is the village map ?

PR #922, opened 2021-04-21
added loadPage Api call

jonnytest1:
allows the APIs to load a new map similar to the exitSceneUrl field 
but contrary to openPage it doesnt reload the entire page or require to rechoose the camera

example usage:
teleporter website where you can choose the map to go to
https://github.com/jonnytest1/workadventuremap/blob/master/scripts/page-picker.js


jonnytest1:
i accidentally added commits to this branch - will redo the pull request

Issue #927, opened 2021-04-21
typing in chat still causes character to move


jonnytest1:
on https://play.workadventu.re/@/tcm/workadventure/demo the character doesnt move ✅ 
but on https://play.staging.workadventu.re/@/tcm/workadventure/demo and my local branch which ive checked is up to date with the develop branch typing in chat causes the character to move as well

now ive only tested it if the chat was opened by 
``` WA.sendChatMessage(text,author)```

jonnytest1:
mediaManager.setUserInputManager(this.userInputManager);
  this.userInputManager = new UserInputManager(this);
  
  am i the only one to whom this sees wrong ? :3

pizkaz:
Sorry, seems I messed this one up while trying to resolve a merge conflict.

jonnytest1:
so um anyone gonna merge this https://github.com/thecodingmachine/workadventure/pull/965 ? ^^

pizkaz:
I included your commit in the [community fork](https://github.com/workadventure-xce/workadventure-xce).

jonnytest1:
> I included your commit in the [community fork](https://github.com/workadventure-xce/workadventure-xce).

Oh didn't know that existed 😅

jonnytest1:
closing since merged

PR #965, opened 2021-04-29
fixed typing in chat still causes character to move

jonnytest1:
userInputManager was always undefined

🤔 there is another bug where if you click on the game canvas the **focus event of the chat input field gets called instead of the blur** ?! so you currently always have to completely clsoe chat to get back into the game


jonnytest1:
fix for https://github.com/thecodingmachine/workadventure/issues/927

jonnytest1:
oh @ClementVieilly75 didnt see you were already assigned

gparant:
Thanks for the fix 💪

PR #970, opened 2021-04-29
The errors messages shown in front

gparant:
- Error link not found
 - Error user was banned
 - Error access dinied on the world

### DIFF
--- a/front/src/Phaser/Login/EntryScene.ts
+++ b/front/src/Phaser/Login/EntryScene.ts
@@ -21,7 +21,10 @@ export class EntryScene extends Scene {
             this.scene.start(nextSceneName);
         }).catch((err) => {
             if (err.response && err.response.status == 404) {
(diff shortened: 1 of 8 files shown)


front/src/Phaser/Login/EntryScene.ts
```diff
@@ -21,7 +21,10 @@ export class EntryScene extends Scene {
             this.scene.start(nextSceneName);
         }).catch((err) => {
             if (err.response && err.response.status == 404) {
-                ErrorScene.showError(new WAError('Page Not Found', 'Could not find map', window.location.pathname), this.scene);
+                ErrorScene.showError(new WAError(
+                    'Access link incorrect', 
+                    'Could not find map. Please check your access link.', 
+                    'If you want more information, you may contact administrator or contact us at: workadventure@thecodingmachine.com'), this.scene);
```
tabascoeye, line 27: Can you turn this into an Environment variable for the self-hosters to change?
You don't want people to spam you with error messages about instances that you don't even host?

Issue #170, opened 2020-06-09
Enhancement : add notification when group is created


nguyenk:
Would be nice if group is created when app is not focused, so the user is aware that someone is trying to reach him.

This will improove experience if there if 'enter' sound could not be heard.

gparant:
@nguyenk can you exeplain in détail your feature ?

VmickaelV:
I don't know if @nguyenk  has the same need as me. But i think we have the same need/idea.
It will be nice, if the app is not focused, when someone wants talk to you, we have a notification. Currently, we have a sound, but some people may have their earphones off, or muted their computer, so we can't ear the sound.
Maybe, some peoples prefer this feature is can be configurable in some settings pannel.

PR #1000, opened 2021-05-06
Notification browser on discussion circle

gparant:
- Notification when a user is first and not focused on the tab.
- The camera focused when user is in discussion circle and back on tab with previous config camera settings.
- The camera stay blurred if user is in discussion circle and not back on the tab.


Kharhamel:
this should fix #170

Kharhamel:
This code is already in develop, so this PR is obsolete

PR #1022, opened 2021-05-10
trigger message api

jonnytest1:
allow triggering of custom triggerMessage


jonnytest1:
@moufmouf

jonnytest1:
@moufmouf @gparant how about it ?

moufmouf (changes requested):
Hey @jonnytest1 ,

Pretty cool.
I'm OK to add this feature to the scripting API, it makes sense.

Also, rather than returning a "string" that can be used to remove the message with `removeTriggerMessage`, I think we should do as we did with Popups: return an object that has a class method:

```
triggerMessage(message:string, callback:() => void): TriggerMessage

class TriggerMessage {
    close(): void;
}
```

Finally, could you add the relevant documentation is the doc? In `docs/maps/api-reference.md`.

Thanks a lot and sorry to take so long to review your PR!

maps/tests/script.js
```diff
@@ -1,40 +1,42 @@
+///<reference path="../../front/src/iframe_api.ts" />
```
jonnytest1, line 1: @moufmouf  this allows type checking in javascript files if checkJs is enabled in tsConfig

front/src/iframe_api.ts
```diff
@@ -1,29 +1,35 @@
...
+    openTab(url: string): void;
+    goToPage(url: string): void;
+    openCoWebSite(url: string): void;
     closeCoWebSite(): void;
     disablePlayerControls(): void;
     restorePlayerControls(): void;
     displayBubble(): void;
     removeBubble(): void;
+
+    triggerMessage(message: string, callback: () => void): {
```
jonnytest1, line 30: i was thinking about extracting it into a class  but honestly i would jsut rather move it to its own file completely 
like described in here

PR #1200, opened 2021-06-17
Add workaround for #932

guysoft:
See https://github.com/thecodingmachine/workadventure/issues/932#issuecomment-830777306

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ Note: on some OSes, you will need to add this line to your `/etc/hosts` file:
 ```
 
 Note: If on the first run you get a page with "network error". Try to ``docker-compose stop`` , then ``docker-compose start``.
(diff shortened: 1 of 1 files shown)


jonnytest1:
sudo docker-compose up --dorce-recreate is what i use

guysoft:
@jonnytest1 Should I change it to that then? Should do the same thing

jonnytest1:
sure 
oh its force-recreate 😅

jonnytest1:
🤔 if you do some sneaky git changes while its running you might need to do it again 
 since the bakcend container can crash if the messages container isnt fast enough in generating the necessary files for compilation - if that happens it ownt even refresh anymore

guysoft:
Sounds to me its best to keep this PR as is since its what I tested and we know that works.  Also when you stop and start you can.confirm the container indeed stopped

PR #1251, opened 2021-06-30
POC for the openID connect

Kharhamel:
We want to completely rewrite the way authentification and authorization work in WA.
The goal of this PR is to create a proof of concept for this:
* We use Google Oauth to allow any user to authenticate on any WA servers (even hosted one).
* We allow pusher servers to create accessToken for authenticated users which can then be used by the ScriptingApi to fetch ressources or push events.
* Allow map creators to associate tags to specifics emails adresses directly in the map metadata, without using the admin plateform.

TODO:

* [x] Implement a login interface
* [x] Implement a refresh mecanism
* [ ] Ban by email/world instead of uuid
* [x] Use a more generic solution than google auth
* [ ] Make the auth jwt token able to be validated by any pusher server (make it signed with an asymetric key)


Lurkars:
Really looking forward to this. Just tested around with own IdP and already works like charm in what's done yet. Are there any plans in forcing OIDC login (so no anonymous login allowed)?

Lurkars:
Also nice feature: set username based on username, preferred_username claim.

Kharhamel:
We don't plan on making mandatory now, not until we are sure it is robust.

PR #1267, opened 2021-07-07
Return at the new line into the Pop-up

gparant:
Add regex to replace "\r\n" or "\r" or "\n" by <br/>


Kharhamel (approved):

Lurkars:
popup-newline
This is not working, because HTML is escaped. I opened a ticket here: #1272

PR #1320, opened 2021-07-28
Load a json file that describe a tileset in Tile format.

GRL78:
Then add it to the tilemap of the GameScene
Then add if to the GameMap


front/src/Phaser/Game/GameScene.ts
```diff
@@ -1081,8 +1082,76 @@ ${escapedMessage}
...
+                    newFirstgid = lastTileset.firstgid + lastTileset.tilecount;
+                }
+                return new Promise((resolve, reject) => {
+                    this.load.on("filecomplete-json-" + eventTileset.url, () => {
+                        let jsonTileset = this.cache.json.get(eventTileset.url);
+                        this.load.image(
+                            jsonTilesetDir + "/" + jsonTileset.image,
+                            jsonTilesetDir + "/" + jsonTileset.image
+                        );
+                        this.load.on("filecomplete-image-" + jsonTilesetDir + "/" + jsonTileset.image, () => {
```
gparant, line 1105: repeat 3 in the code, you can create const of `jsonTilesetDir + "/" + jsonTileset.image`

PR #1369, opened 2021-08-13
Exit scene acess denied detected

gparant:
Permit to check access on "exitUrl" properties. Il user haven't access we show error message


jonnytest1:
dont forget to implement the denial basckend side as well - its pretty easy to circumvent frontend permission checks

gparant:
@jonnytest1 thanks for your feedback, don't worry the security check was developed in our Back Office 🚀

Moini:
I get this error when the map doesn't load for some reason, even though it should be working. Possibly to do with change to exitUrl.

Moini:
(btw. suggesting to fix the English grammar on the denial sentence: "Room access denied. You don't have permission to access this room.")

PR #1377, opened 2021-08-16
Active authentication Oauth

gparant:
- Google authentication
 - GitHub authentication
 - Linkedin authentication


pusher/src/Controller/AuthenticateController.ts
```diff
@@ -48,21 +49,62 @@ export class AuthenticateController extends BaseController {
...
                 res.writeStatus("200");
                 this.addCorsHeaders(res);
                 return res.end(JSON.stringify({ authToken }));
             } catch (e) {
                 return this.errorToResponse(e, res);
             }
         });
+
+        // eslint-disable-next-line @typescript-eslint/no-misused-promises
+        this.App.get("/logout-callback", async (res: HttpResponse, req: HttpRequest) => {
```
Kharhamel, line 86: why do you need a logout callback? Simply deleting the token is enough

front/src/Connexion/LocalUserStore.ts
```diff
@@ -160,19 +164,24 @@ class LocalUserStore {
 
     verifyState(value: string): boolean {
         const oldValue = localStorage.getItem(state);
-        localStorage.removeItem(state);
         return oldValue === value;
     }
+    getState(): string | null {
```
Kharhamel, line 169: state and nonce should never be reused. there are supposed to be newly created at each login attempt.

front/src/Connexion/LocalUserStore.ts
```diff
@@ -160,19 +164,24 @@ class LocalUserStore {
...
     getNonce(): string | null {
-        const oldValue = localStorage.getItem(nonce);
-        localStorage.removeItem(nonce);
-        return oldValue;
+        return localStorage.getItem(nonce);
+    }
+    setCode(value: string): void {
+        localStorage.setItem(code, value);
+    }
+    getCode(): string | null {
```
Kharhamel, line 183: the auth code should not be saved. It is supposed to be a temporary ressource used to get tokens

Kharhamel (commented):
Don't forget to delete gameMenu.html

pusher/src/Services/AdminApi.ts
```diff
@@ -141,6 +141,15 @@ class AdminApi {
             return data.data;
         });
     }
+
+    /*TODO add constant to use profile companny*/
+    getProfileUrl(accessToken: string): string {
+        if (!ADMIN_URL) {
```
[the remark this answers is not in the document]
gparant: Because the PROFILE_URL is only use in front. Here we defined our profile URL

pusher/src/Controller/AuthenticateController.ts
gparant: Logout callback permit to defined logout user in hydra

front/src/Connexion/LocalUserStore.ts
gparant: Yes, I know it will be change

front/src/Connexion/LocalUserStore.ts
gparant: Yes, I know it will be change

PR #1399, opened 2021-08-31
Profile url

gparant:
Create profile URL for OpenId Auth

### DIFF
--- a/front/src/Connexion/LocalUserStore.ts
+++ b/front/src/Connexion/LocalUserStore.ts
@@ -1,5 +1,6 @@
 import { areCharacterLayersValid, isUserNameValid, LocalUser } from "./LocalUser";
 import { v4 as uuidv4 } from "uuid";
+import { START_ROOM_URL } from "../Enum/EnvironmentVariable";
 
 const playerNameKey = "playerName";
 const selectedPlayerKey = "selectedPlayer";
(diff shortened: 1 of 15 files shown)


front/src/Connexion/LocalUserStore.ts
```diff
@@ -126,7 +127,9 @@ class LocalUserStore {
         });
     }
     getLastRoomUrl(): string {
-        return localStorage.getItem(lastRoomUrl) ?? "";
+        return (
+            localStorage.getItem(lastRoomUrl) ?? window.location.protocol + "//" + window.location.host + START_ROOM_URL
```
jonnytest1, line 131: protocol+ host = location.origin

pusher/src/Services/AdminApi.ts
```diff
@@ -141,6 +141,15 @@ class AdminApi {
             return data.data;
         });
     }
+
+    /*TODO add constant to use profile companny*/
+    getProfileUrl(accessToken: string): string {
+        if (!ADMIN_URL) {
```
tabascoeye, line 147: wait. ADMIN_URL is the deprecated version of API_URL.
I think _here_ you mean ADMIN_API_URL...

pusher/src/Services/AdminApi.ts
```diff
@@ -141,6 +141,15 @@ class AdminApi {
...
+            throw new Error("No admin backoffice set!");
+        }
+
+        return ADMIN_URL + `/profile?token=${accessToken}`;
```
tabascoeye, line 151: ...and here you probably mean API_URL if it should be sent to the pusher first?!

PR #1405, opened 2021-09-01
openTabPropertyKey (create new props in own file)

jonnytest1:


jonnytest1:
also fixed any jitsiTypes

front/src/Phaser/Game/GameMapPropertyChange.ts
```diff
@@ -0,0 +1,55 @@
...
+import {
+    ON_ACTION_TRIGGER_BUTTON,
+    TRIGGER_WEBSITE_PROPERTIES,
+    WEBSITE_MESSAGE_PROPERTIES,
+} from "../../WebRtc/LayoutManager";
+
+export class GameMapPropertyChange {
+    constructor(private scene: GameScene, private gameMap: GameMap) {}
+
+    register() {
```
Kharhamel, line 15: you might as well just put all this code directly into the constructor

front/src/Phaser/Game/GameMapPropertyChange.ts
```diff
@@ -0,0 +1,55 @@
...
+        this.gameMap.onPropertyChange("openTab", (newValue) => {
+            if (typeof newValue == "string" && newValue.length) {
+                scriptUtils.openTab(newValue);
+            }
+        });
+        this.gameMap.onPropertyChange("openWebsite", (newValue, oldValue, allProps) => {
```
Kharhamel, line 21: just to be sure, you didn't edit this.gameMap.onPropertyChange("openWebsite", (...)) ?
You simply moved it away from GameScene?

front/src/Phaser/Game/GameMapPropertyChange.ts
```diff
@@ -0,0 +1,55 @@
...
+import { scriptUtils } from "../../Api/ScriptUtils";
+import { coWebsiteManager } from "../../WebRtc/CoWebsiteManager";
+import { layoutManagerActionStore } from "../../Stores/LayoutManagerStore";
+import {
+    ON_ACTION_TRIGGER_BUTTON,
+    TRIGGER_WEBSITE_PROPERTIES,
+    WEBSITE_MESSAGE_PROPERTIES,
+} from "../../WebRtc/LayoutManager";
+
+export class GameMapPropertyChange {
```
Kharhamel, line 12: I would rather call this class GameMapPropertiesListener

front/src/Phaser/Game/GameMapPropertyChange.ts
Kharhamel: Also you could put all the other this.gameMap.onPropertyChange from GameScene here, and maybe put every listened properties into constants, that way we would have an easy to list map properties

jonnytest1: well if i did that we'd have to make some properties public that were private before 
personally id also prefer if the implementation of the listeners was somewhere external - to basically have a single easily graspable file that just has one line for each listener

front/src/Phaser/Game/GameMapPropertyChange.ts
jonnytest1: well its is slightly adjusted - i converted 2 `this.` to `this.scene.`

front/src/Phaser/Game/GameMapPropertyChange.ts
jonnytest1: i dont think typescript likes just calling new () without then using the returned value

jonnytest1: how about we just convert the entire calss to a function ?

jonnytest1: but the gameScene class is already waaaay too big and refactorings will usally only be done when christmas is on the same day as easter (never :D )

Kharhamel: But I like to do big refactorings :( . But you're not wrong and the GameScene is very tricky to edit, I am just not a that to create random class that don't really have a well contained purpose. I'll guess I'll allow it this time because this PR is usefull

PR #1431, opened 2021-09-09
FIX: the video element should not have a bigger height than its container

Kharhamel:


moufmouf:
Almost!
We have an issue when enabling going in "presentation" mode now:

image

PR #1434, opened 2021-09-09
explain new property jitsiWidth

tabascoeye:


Kharhamel:
Good catch. This must have been forgotten when @moufmouf migrated the doc

PR #1301, opened 2021-07-21
Twemoji emote menu

Lurkars:
This uses Twemoji Font to show Emotes and an Emoji Picker (emoji-button) for selecting emotes.

The positioning on initial opening is not ideal, I am not deep enough into phaser to understand what's going on there, so will need some tweaks, but in general I think it is usable and a good improvement.

PR #1437, opened 2021-09-10
Twemoji emote menu svelte (upgrade of #1301)

Lurkars:
I migrated my PR #1301 to a solution in svelte. This solves the mobile issues and is hopefully the better solution, then rendering HTML in Phaser. I am a bit unsure if I used the svelte store as intended.

Please have a look @moufmouf and if you prefer svelte solution #1301 can be closed than in favor of this.


Lurkars:
Here little demo video:
https://user-images.githubusercontent.com/2760570/132878384-e0f96147-8b14-4191-a3d3-665b97647a1f.mp4

front/src/Phaser/Game/GameScene.ts
```diff
@@ -614,6 +616,15 @@ export class GameScene extends DirtyScene {
             this.openChatIcon.setVisible(!v);
         });
 
+        this.emoteUnsubscribe = emoteStore.subscribe(() => {
```
Kharhamel, line 619: When you subscribe, the first parameter of the callback is the new value. So you can write:
```
.subscribe((emoteKey) => {
            if (emoteKey) {
                ...
                }
                });

File: front/src/Phaser/Player/Player.ts [hunk 2]
```diff
@@ -94,39 +85,16 @@ export class Player extends Character {
...
-        this.emoteMenu = null;
+    playEmote(emote: string) {
+        super.playEmote(emote);
+        emoteMenuStore.set(false);
     }
 
-    destroy() {
-        this.scene.events.removeListener("postupdate", this.updateListener);
-        super.destroy();
+    openOrCloseEmoteMenu() {
```
Kharhamel, line 93: I would take this method out of the class Player since it doesn't really interact with the phaser object. You could put it directly inside the emoteMenuStore or at least in EmoteStore.ts

Kharhamel:
I created another PR so I could run a deploy, you can test it here: https://play-twemojiemotemenusvelte.test.workadventu.re

This looks really good! Maybe change some style on the menu to look more like other menus, but the base code is good. I never tough about using texts instead of sprites.

Kharhamel:
Also, I think having the searchbar would be better.

Lurkars:
> Also, I think having the searchbar would be better. 

The reason I disabled it on purpose is actually the player movement. First of all, I really like to not stop moving while having emoji picker open and otherwise player controls need to be disabled due to WASD keys when typing.

Kharhamel:
We could try to disable movment keys only when you focus the searchbar, does the searchbar have a id so we can get its DOM easily?

front/src/Phaser/Player/Player.ts
Lurkars: yes make sense, was just adopted from old code.

Lurkars:
> We could try to disable movment keys only when you focus the searchbar, does the searchbar have a id so we can get its DOM easily?

The search has a class and can be easily referenced. The Problem is, that the whole picker listens to a keydown event and then focus the search if present (see https://github.com/joeattardi/emoji-button/blob/d859569bb3d9d9180c2479699396e0ec5bc7b899/src/index.ts#L718)

So it's getting a bit complicated, because if you just listen to focus event, it's just gets triggered directly when typing and would disable controls immediately. So in the end the behavior would be complete same, as if controls just get directly disabled when opening menu.

Kharhamel:
Yeah i think we can forget the searchbar for now. But the menu needs to be fixed on mobile, it is wider than the screen

GRL78:
- The emote menu doesn't work when the woka is in a jitsi room. 
- Every time, i open the emote menu WA freeze.
- Need to block control of the player when emotes menu is open, because the arrow keys can be use to navigate between the emote.

Lurkars:
I fixed the Jitsi stuff, can you describe more what you mean with WA freezes completly? I do not notice this issue, but I only tested firefox and chrome on ubuntu.

Oh yes, with the controls I didn't notice while always using WASD. So I think we should just disable controls when picker is open, then the search bar can also be displayed...

Lurkars:
@Kharhamel I added some improvements:
- own store for svelte to fix jitsiMeet issued
- disabled controls on open (and adding search bar again)

 so you may want to merge this into your PR. About the mobile issues: on my test the picker is round about 380px wide and I had no problems on mobile. Of course I cannot test on smaller phones, but I think we would just need then a different configuration for EmojiButton on mobile. You can play around with parameters like "emojiSize" and "emojisPerRow", maybe also changing "showCategoryButtons" as the documentation suggests ( https://emoji-button.js.org/docs/api )

Kharhamel:
@Lurkars I updated my MR to include your recent changes and I also pushed two new changes:
 * Some new style to be more consistent with other menus
 * Only 6 columns on mobile

You can check the deploy https://play-twemojiemotemenusvelte.test.workadventu.re to see if the new style suits you or I you have another idea.

I think the PR is mergable as it is. Could you simply fix your conflicts? You can integrate my changes in your own commits if you want, or just update your own PR and then I'll update my own

Lurkars:
I find with the changes. Hope the merges are correct now for you.

Kharhamel:
Since you integrate my changes, I will be merging this PR and closing mine. Thank you for the feature!

PR #1438, opened 2021-09-10
Twemoji emote menu svelte

Kharhamel:
This PR is a duplicate of #1437
This is the only I found to run a deploy on this feature


GRL78:
> The emote menu doesn't work when the woka is in a jitsi room. 
> Every time, i open the emote menu WA freeze.
> Need to block control of the player when emotes menu is open, because the arrow keys can be use to navigate between the emote.

Kharhamel:
Closed since the base PR was merged

PR #1453, opened 2021-09-14
WIP : Change bitrate of peerConnection in settings menu

GRL78:


GRL78:
This PR doesn't use the right way to control quality of video and audio stream of the user. Use videoConstraintStore and AudioConstraintStore instead.

PR #1480, opened 2021-09-23
Removed old maps and add starter-kit map

ValdoTR:
I've removed all deprecated code inside the */maps* folder, and the old maps also.
Now the default map is */maps/starter/map.json* which is the same map as the starter-kit repo, with the WA tilesets.

I did not touch the */maps/tests* folder (mandatory for our integration tests) and I have kept the */maps/Tuto* for a future upgrade.


Kharhamel:
Can you edit START_ROOM_URL in .env.template to reference the new start map?

ValdoTR:
Oh my bad... I'll do it

PR #1489, opened 2021-09-27
Fix : Remove audio manager range volume control by keys

nolway:


front/src/Components/AudioManager/AudioManager.svelte
```diff
@@ -88,6 +95,11 @@
         changeVolume();
     }
 
+    function disallowKeys() {
+        audioPlayerVol.blur();
+        return false;
```
Kharhamel, line 100: why the return false?

PR #1521, opened 2021-10-14
Change character depth to 0

nolway:
Currently the characters can be display under a text or something object with a depth of -1.

WARNING : Test your maps to abort deviant behaviors.


gparant:
@moufmouf could merge it?

PR #1533, opened 2021-10-21
use OIDC without admin api, option to disable anonymous login

Lurkars:
This PR as two main focuses:

- First of all it makes OpenID usable without Admin API by using the Pusher instead of `iframeAuthentication` when not set (which can normally only be set with Admin API). 
- Seconds it adds an `DISABLE_ANONYMOUS` option, which of course only makes sense with usage of OpenID with following things happening if disabled
    -  "/anonymLogin" returns 403 instead of a token
    - all rooms gets `authenticationMandatory` set to true
    - "/room" needs a valid token (this is necessary to avoid hacks, if OpenID redirect is omitted somehow, it would be possible to set username etc and to proceed to map without an authToken set!!!)

This is both in a single PR, because using OpenID without Admin API does not really make sense without `DISABLE_ANONYMOUS`. So I think this belongs together!

### DIFF
--- a/pusher/src/Controller/IoSocketController.ts
+++ b/pusher/src/Controller/IoSocketController.ts
@@ -26,7 +26,7 @@ import { jwtTokenManager, tokenInvalidException } from "../Services/JWTTokenMana
 import { adminApi, FetchMemberDataByUuidResponse } from "../Services/AdminApi";
 import { SocketManager, socketManager } from "../Services/SocketManager";
 import { emitInBatch } from "../Services/IoSocketHelpers";
-import { ADMIN_API_TOKEN, ADMIN_API_URL, SOCKET_IDLE_TIMER } from "../Enum/EnvironmentVariable";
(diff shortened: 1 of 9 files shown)


pusher/src/Controller/IoSocketController.ts
```diff
@@ -175,6 +175,11 @@ export class IoSocketController {
 
                         const tokenData =
                             token && typeof token === "string" ? jwtTokenManager.verifyJWTToken(token) : null;
+
+                        if (DISABLE_ANONYMOUS && !tokenData) {
```
gparant, line 179: Here we evaluate sokect connexion and the disable anonymous can be checked here

gparant:
Hello @Lurkars, I'm back with documentation (schema) of OpenId.

WA - OpenId connection self hosted

Some explains: 
 - The green is front and pusher,
 - The red is OpenId provider, WorkAdventure use hydra to have OpenId protocol,
 - The Yellow is the login / consent and logout provider. WorkAdventure use a private provider developed in our back office.

Don't hesitate if you have any question or let me a feedback 💪 🚀

PR #1554, opened 2021-11-10
Implement Twemoji on emote

nolway:
Replace the current emote text format to the Twemoji svg library.


Lurkars:
I like the change to SVG, but now we again relying on third party: twemoji.maxcdn.com so I would recommend to also ship the svgs within WA! @moufmouf

PR #1559, opened 2021-11-12
Open id disabled anonymous self hosted

gparant:


gparant:
Crteate simple profile url in pusher container

https://user-images.githubusercontent.com/23724509/141526345-590c6b6f-5062-4d30-91f6-b1fb95a8c7ae.mov

PR #1623, opened 2021-12-06
Feature: Following WOKAs

pizkaz:
This is a "preview" of a WIP feature we are implementing: You can now press the interact button (`[space]`) to start following the nearest WOKA. It is still quite rough but usable.

So far we only collected the low-hanging fruits: It runs completely without client/server communication but improving it further will probably require changes to the `front <-> pusher` protocol, so before we start digging into that I would appreciate some feedback:
* Is such a feature of interest to anyone?
* Would maintainers be willing to merge it? The last thing we want is a "hostile" fork.

Things that definitely could be better:
* Ongoing conversations should be protected when following. Currently, conversations are closed if the "leading" WOKA moves too fast.
* The frontend should clearly display that follow mode is enabled - preferably on both sides (following / followed).
* (not sure if this is necessary:) The player being followed should be asked if it is OK and have the opportunity to deny the request.
* Conversations should persist beyond the edges of a map (=> `exitUrl`).

### DIFF
--- a/front/src/Connexion/ConnectionManager.ts
+++ b/front/src/Connexion/ConnectionManager.ts
@@ -1,5 +1,5 @@
 import Axios from "axios";
-import { PUSHER_URL, START_ROOM_URL } from "../Enum/EnvironmentVariable";
+import { PUSHER_URL } from "../Enum/EnvironmentVariable";
 import { RoomConnection } from "./RoomConnection";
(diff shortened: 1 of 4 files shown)


pizkaz:
Meh, didn't know that renaming branches closes pull-requests.

=> Moved to #1627

PR #1626, opened 2021-12-06
Display link to map in map credits if provided in map properties

pizkaz:
This patch incorporates a "new" map property `mapLink`, which is rendered as a hyperlink in map credits. It is not really a new map property because it is already present in test maps.

I also found that another property `mapImage`, which is also included in test maps, is not rendered either.

### DIFF
--- a/front/src/Components/Menu/AboutRoomSubMenu.svelte
+++ b/front/src/Components/Menu/AboutRoomSubMenu.svelte
@@ -8,6 +8,7 @@
     let expandedTilesetCopyright = false;
 
     let mapName: string = "";
+    let mapLink: string = "";
     let mapDescription: string = "";
     let mapCopyright: string = "The map creator did not declare a copyright for the map.";
     let tilesetCopyright: string[] = [];
@@ -17,6 +18,10 @@
             const propertyName = gameScene.mapFile.properties.find((property) => property.name === "mapName");
             if (propertyName !== undefined && typeof propertyName.value === "string") {
                 mapName = propertyName.value;
+            }
+            const propertyLink = gameScene.mapFile.properties.find((property) => property.name === "mapLink");
+            if (propertyLink !== undefined && typeof propertyLink.value === "string") {
+                mapLink = propertyLink.value;
             }
             const propertyDescription = gameScene.mapFile.properties.find(
                 (property) => property.name === "mapDescription"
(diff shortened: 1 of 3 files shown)


front/src/Components/Menu/AboutRoomSubMenu.svelte
```diff
@@ -48,6 +53,9 @@
     <section class="container-overflow">
         <h3>{mapName}</h3>
         <p class="string-HTML">{mapDescription}</p>
+        {#if mapLink}
+            <p>&gt; <a href={mapLink}>link to this map</a> &lt;</p>
```
moufmouf, line 57: Maybe the link should open a new tab:

```suggestion
            <p>&gt; <a href={mapLink} target="_blank">link to this map</a> &lt;</p>
```

moufmouf:
Hey @pizkaz ,

Looking at the code, I realize there are a number of properties we never really documented:

- mapName
- mapDescription
- mapCopyright
- tilesetCopyright
- mapLink

We use those in the website to display what map looks like. For instance here: https://workadventu.re/getting-started?mapUrl=https://gparant.github.io/medium-webinar/map.json

The `mapLink` should be used to point to the Github directory, and **not** the map file itself as we did in some places which makes no sense (if we managed to download the map file, we already new the map URL ;) )

I'm :+1: merging this but you should first make the link open in a new tab.

front/src/Components/Menu/AboutRoomSubMenu.svelte
pizkaz: Yeah, already thought about that myself. Will change and rebase it. :-)

pizkaz:
> Looking at the code, I realize there are a number of properties we never really documented:
> 
>     * mapName
>     * mapDescription
>     * mapCopyright
>     * tilesetCopyright
>     * mapLink

Yep, all of these will be used for rC3 maps. That's how we stumbled upon `mapLink` not being implemented. ;-)

There's also `mapImage` but (at least for now) we can't imagine any real use for that.

front/src/Components/Menu/AboutRoomSubMenu.svelte
pizkaz: Done. I also added `class="string-HTML"` to the surrounding `<p>`, because the other `<p>`s also have it.
And `"..."` around the actual URL in `href=`.

tabascoeye:
I suspect the purpose of `mapInage` was in the admin backend? @moufmouf

Screenshot 2021-12-08 at 15 57 34

tabascoeye:
TODO: add `mapLink`to docs

pizkaz:
I added documentation on `mapLink` and a new layer property `audioCopyright` including an example in the properties test map and documentation.

pizkaz:
For clarity: I think, this is ready to be merged.
@moufmouf

moufmouf:
Thanks a lot!

PR #1672, opened 2021-12-23
2 days limit

gparant:
Capture d’écran 2021-12-23 à 12 36 59

- [x] Add warning in WorkAdventure
- [x] Add limit for a map
- [x] Add modal register
- [x] Add modal invit user
- [x] Change menu to add invit user and register


front/src/Components/WarningContainer/WarningContainer.svelte
```diff
@@ -1,20 +1,28 @@
...
     {#if $userIsAdminStore}
+        <h2>Warning!</h2>
         <p>
             This world is close to its limit!. You can upgrade its capacity <a href={upgradeLink} target="_blank"
                 >here</a
             >
         </p>
+    {:else if $limitMapStore}
+        <p>
+            Your are une test mode. This map will be opened during 2 days. You can register your domain <a
```
tabascoeye, line 20: s/une/in/ I guess?
Also "This map is available for 2 days." makes more sense, I would say

PR #1691, opened 2021-12-31
[rc3] Various settings

pizkaz:
Implements various settings from rC3:

* Disabling the camera
* Silent mode (silent zone everywhere)
* Ignoring `playAudio` properties
* Disabling tile animations
* Forcing the Jitsi-/cowebsite trigger (you *always* get asked before opening a cowebsite)
* Blocking external content entirely (no cowebsites, Jitsis, tabs - you don't even get asked)

Supersedes #1678 

Note: Drops legacy map property `playAudioLoop`!

### DIFF
--- a/play/src/front/Components/Menu/SettingsSubMenu.svelte
+++ b/play/src/front/Components/Menu/SettingsSubMenu.svelte
@@ -11,6 +11,7 @@
     import { isMediaBreakpointUp } from "../../Utils/BreakpointsUtils";
     import { audioManagerVolumeStore } from "../../Stores/AudioManagerStore";
     import { onMount } from "svelte";
+    import { gameManager } from "../../Phaser/Game/GameManager";
 
     import infoImg from "../images/info.svg";
     import { iframeListener } from "../../Api/IframeListener";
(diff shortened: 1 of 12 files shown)


pizkaz:
Rebased again. :-)

pizkaz:
Meh, probably conflicts with #1945 now.

I could rebase this again but would like some directions before doing that:

* First of all: Do we want all of these options? If not: Which ones do we want?
  Personally, I would really like to see "Disable camera", "Silent mode" and "Force co-website triggers onaction" merged to upstream. The others I could do without and/or keep those as patches in XCE (until I get tired of rebasing them ;-).
* We have quite a lot of checkboxes in the settings menu already and this PR adds up to six more.
  I guess it's time to start talking about UX. Maybe we should group options?

pizkaz:
As discussed yesterday, this PR adds two settings for:

* disabling tile animations
* disabling proximity chat/meetings

What I overlooked yesterday is another setting which blocks audio tiles. If you do not want to merge this, please let me know, so I can remove the commit from this PR.

**Warning**:
There is also a commit in here which drops the legacy `playAudioLoop` map layer property.

pizkaz:
Oh yes, oh yes. I'm back!

In the meantime the `playAudioLoop` map property was dropped already.
But there are new i18n strings in languages I don't speak. How should I proceed regarding those? Google translate? ;-)

Issue #1577, opened 2021-11-22
Feature Request: player.isInZone


Lithimlin:
Such a function, determining whether a player is inside a zone, would be very helpful for scripting, for example for determining where a player spawned.

Jacques-Olivier-Farcy:
Hi @Lithimlin ,

Nice idea as the WA.player.onPlayerMove(console.log); function only shown coordinate.

nolway:
Hi @Lithimlin ,
We have a subscriber that fires when a user enters or leaves a map layer 😀
https://workadventu.re/map-building/api-room.md#detecting-when-the-user-entersleaves-a-layer

Is this what you are looking for?

Lithimlin:
Hi @nolway,
no, this is not what I am looking for. I am already aware of those.
I'm specifically interested in a way to check whether or not a player is already inside of a zone/layer. For example, when the player spawns, they do not trigger `onLayerEnter()`. It is thus not helpful in determining where a player spawns. `onPlayerMove()` is similarly useless until a player moves and much more complicated.

gparant:
Hi @Lithimlin I think that we will fix the problem of onLayerEnter when user spawns, but with function onEnterLayer and onLeaveLayer, you could define if the user is in layer or you could go in the future and define global variable to stock the route of all users and use it in your script.

I will add this fix in our issue. Thanks for your feedbakc 💪 🚀

PR #1725, opened 2022-01-13
[Draft] Implement spawnLayers property on iframe api

nolway:
Add spawn layers on iframe api : https://github.com/thecodingmachine/workadventure/issues/1577

### DIFF
--- a/front/src/Api/Events/GameStateEvent.ts
+++ b/front/src/Api/Events/GameStateEvent.ts
@@ -6,7 +6,7 @@ export const isGameStateEvent = new tg.IsInterface()
         mapUrl: tg.isString,
         nickname: tg.isString,
         uuid: tg.isUnion(tg.isString, tg.isUndefined),
-        startLayerName: tg.isUnion(tg.isString, tg.isNull),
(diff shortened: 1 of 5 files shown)


moufmouf (changes requested):
To be honest, I'm not a fan of the idea of adding a "spawnLayers" property.

This is a property that is only valid when the user arrives and that is useless afterwards. Still, it is stored forever here.

Furthermore, it does not fully answer #1577. Indeed, if `onEnterLayer` is not called when the user enters the room but afterwards, we have no way to know what layers the user is on if he has moved.

I would really favor one of these 2 alternative solutions.

**Solution 1:**

This is something we can do in "scripting API extra", by reading the map, getting the user coordinates, and finding all the layers of the user (and this is more useful because if we do it this way, we can get the list of layer of any (x, y) coordinates and not the layers of the spawn point.

**Solution 2:**

`onEnterLayer` could trigger the callback automatically if the user is already in the given layer.
Or alternatively, we could add an additional "onStart" parameter that triggers the callback if the user is already on the layer when the `onEnterLayer` method is called.

PR #1756, opened 2022-01-25
Enhance emoji menu

nolway:
Enhance emoji menu design & add all sentences use into the translation system.

Screenshot from 2022-01-26 10-40-33


front/src/Components/EmoteMenu/EmoteMenu.svelte
```diff
@@ -15,10 +16,31 @@
             rootElement: emojiContainer,
             styleProperties: {
                 "--font": "Press Start 2P",
+                "--text-color": "whitesmoke",
```
gparant, line 19: What do you use `whitesmoke` color ?

nolway: Whitesmoke is used on many parts of the project and the previous colour of the text does not have sufficient contrast to be seen by the colour blind.

PR #1759, opened 2022-01-25
Fix stacking website triggers

### LINKED ISSUE #1758: Stacking open-website triggers
Steps to reproduce:

1. set `forceCowebsiteTrigger` to `true`
1. walk in and out of a website area multiple times
1. :boom: the website trigger is not removed on walking out and gets added again when moving back into the area
1. when now pressing _Space_ the website will open multiple times

lukashass:
Closes #1758


lukashass:
This might no longer be relevant, since #1620 replaced the message popup with a coffee mug :sweat_smile:

PR #1760, opened 2022-01-26
de-DE german translation (+ small improvement)

Lurkars:
Here is a basic translation for german. 
Also during translation I recognized, that the video quality has a "small" option which is smaller than "minimum". From my logic point of view, this should be changed, because there isn't a smaller value than minimum. So I included this in this PR. Hope this is okay for you.


nolway:
Thank you for your contribution !
Can you add the new translations of the emoji menu ? : https://github.com/thecodingmachine/workadventure/pull/1756

You also need to run Prettier on the front part to pass the continuous integration job =)

PR #1820, opened 2022-02-03
Feature voice chat indicator

Waclaw-I:
Cute dialog-clouds appear above players heads when talking in circle.
obraz


front/src/Components/EnableCamera/EnableCameraScene.svelte
```diff
@@ -86,7 +95,7 @@
             <img class="background-img" src={cinemaCloseImg} alt="" />
         </div>
     {/if}
-    <HorizontalSoundMeterWidget {stream} />
+    <HorizontalSoundMeterWidget {volume} />
```
moufmouf, line 98: You can do even better here (I think).

In a Svelte template, you can use the value of store directly by prefixing it with a "$".

So instead of subscribing / unsubscribing to the store, you can write:

```suggestion
    <HorizontalSoundMeterWidget volume={$localVolumeStore} />
```

front/src/Components/MyCamera.svelte
```diff
@@ -51,7 +59,7 @@
         <div class="is-silent">{$LL.camera.my.silentZone()}</div>
     {:else if $localStreamStore.type === "success" && $localStreamStore.stream}
         <video class="my-cam-video" use:srcObject={stream} autoplay muted playsinline />
-        <SoundMeterWidget {stream} />
+        <SoundMeterWidget {volume} />
```
moufmouf, line 62: Same comment here, you can use the store directly.

front/src/Components/Video/VideoMediaBox.svelte
```diff
@@ -93,7 +103,7 @@
     />
     <img src={blockSignImg} draggable="false" on:dragstart|preventDefault={noDrag} class="block-logo" alt="Block" />
     {#if $constraintStore && $constraintStore.audio !== false}
-        <SoundMeterWidget stream={$streamStore} />
+        <SoundMeterWidget {volume} />
```
moufmouf, line 106: Same here, try to use $volumeStore directly.

front/src/WebRtc/VideoPeer.ts
```diff
@@ -69,6 +73,35 @@ export class VideoPeer extends Peer {
...
+                        }
+                    }
+                }, 100);
+            });
+
+            return () => {
+                unsubscribe();
+                clearInterval(timeout);
+            };
+        });
```
moufmouf, line 103: This code is very similar to the one defining the localVolumeStore.

Could we maybe try to make a function that acts as a factory of volume stores and that is used in both places?

Something like:

```
export function createVolumeStore(streamStore: Readable<Stream>): Readable<number | null> {
   // ...
}
```

moufmouf (commented):
I love how this PR is actually making the code more elegant in the Svelte components!

I added a few comments regarding things we can probably simplify a bit (using the "$" store notation of Svelte)

moufmouf:
A few comments:

- [ ] It doesn't seem to work on Firefox. I see the indicator on Chrome but on Firefox, I don't see mine, nor the one from other users (which is surprising because I do see the indicator in the video)
- [ ] The indicator is maybe a bit too sensitive. On my microphone which has just a little bit of background noise, the indicator is always on (and I'd probably like it always off)
- [ ] When I am in a bubble and I exit the bubble, if I keep talking, I see the indicator. I expect not to see it (since noone can here me, nor see my indicator)

Waclaw-I:
> A few comments:
>     * [ ]  When I am in a bubble and I exit the bubble, if I keep talking, I see the indicator. I expect not to see it (since noone can here me, nor see my indicator)

I cannot reproduce this behaviour on FF, Chrome and Brave. Is it still occuring?

moufmouf:
I think a file is missing in the latest commit:

image

(happens in both Chrome and Firefox)

moufmouf:
> I cannot reproduce this behaviour on FF, Chrome and Brave. Is it still occuring?

Alas yes.
We managed to reproduce this on Chrome and Firefox:

https://user-images.githubusercontent.com/1290952/154270061-07c912f4-da06-4f49-ba00-efd272ba98e7.mp4

moufmouf:
It seems to be easily reproducible when there are 3 people in a bubble.

moufmouf:
Yeah! :tada:

PR #1990, opened 2022-03-21
Allow audio feature

ValdoTR:
Mainly to be able to play sounds in mobile, I have reworked a little bit the audio manager.
Now we have a "Allow audio" button if the sound is not playing right away.


ValdoTR:
Closing this to push on upstream instead because of the deploy label

PR #1994, opened 2022-03-22
New custom woka scene

Waclaw-I:
https://user-images.githubusercontent.com/14911927/159500057-54489224-e114-43fb-8669-004932ef0eaf.mp4


moufmouf:
Could you change the font of the text on the buttons so that it matches the font used for other buttons?

front/yarn.lock
```diff
@@ -2213,6 +2220,14 @@ phaser3-rex-plugins@^1.1.42:
...
     webfontloader "^1.6.28"
 
+phaser@3.55.1:
+  version "3.55.1"
+  resolved "https://registry.yarnpkg.com/phaser/-/phaser-3.55.1.tgz#25923fe845f6598aec57cfb37a5641834e9943a7"
+  integrity sha512-A5J9/diRz05qc498UNJAaXp85JVkBAEMqxP8pmcRMu1RCLBs4Kx7axd7YxNbXnQuK58JBhTRucngLt8LSpsUlQ==
+  dependencies:
+    eventemitter3 "^4.0.7"
+    path "^0.12.7"
+
```
moufmouf, line 2230: Weird... why do we have 2 phaser versions in yarn.lock?

moufmouf:
When we arrive on the customize page, the grid at the bottom of the screen is empty:

image

It would make sense to load with body textures I think.

moufmouf:
I think I would change the order of the "parts" buttons (to order them from most important to less important).

image

So on the left, I'd put (from top to bottom):

- body color
- eyes
- hairs

And to the right (from top to bottom)

- clothes
- hats
- accessories

HimeShaman:
Capture d’écran du 2022-03-24 14-38-17

Maybe we could add arrows on right & left end of the bottom grid in order to let user know that they can scroll through a list of option (I tried some opacity effect but arrows only could probably work) 

Also some visual hint could help people know than the woka can be shown in another angle on clic.

ValdoTR:
The whole thing seems a little bit too big and "impressive" the first time you see it. Maybe you should give the borders more space (margin in the container)

Waclaw-I:
> 

I think it would spark situations where you are unable to see some parts because they were hidden by something already equipped, with higher render priority (hat covering different hairstyles) thus forcing user to first unequip parts in order to see clearly.
Moreover, it would be inconsistent with selecting BodyType - Here I am almost sure we should leave just bodies to be shown as they sometimes differ in very subtle details.

moufmouf:
A few additional feedbacks:

- [ ] When clicking on one of the 6 "parts" button, the list at the bottom of the screen does not automatically center on the currently selected part. Furthermore, the currently selected part is not highlighted
- [ ] Comment from @gparant (I did the same a few days ago): could we try to display the currently selected woka (with all parts except the part we are selecting) in the grid at the bottom of the screen. So for instance, if I select a phantom body with pink eyes and a sword, when I choose a hat, all the hats in the grid are shown on top of a phantom body with pink eyes with a sword
- [ ] For the Woka preview, instead of displaying white background, could you try to put it on top of a typical "floor" background? (like some kind of grass or rock). This would help understand that some bodies are actually transparent (ghooOOoooOOoost :ghost: )
- [ ] Question from @gparant : Could you try a black background with white borders (instead of white background with grey borders) ? => try to put this in a separate commit so we can revert it if this is a bad idea :D

front/src/Phaser/Login/EntryScene.ts
```diff
@@ -46,7 +47,8 @@ export class EntryScene extends Scene {
                         // Let's rescale before starting the game
                         // We can do it at this stage.
                         waScaleManager.applyNewSize();
-                        this.scene.start(nextSceneName);
+                        // this.scene.start(nextSceneName);
+                        this.scene.start(CustomizeSceneName);
```
moufmouf, line 51:
```suggestion
                        this.scene.start(nextSceneName);
```

front/src/Phaser/Login/EntryScene.ts
```diff
@@ -9,6 +9,7 @@ import { get } from "svelte/store";
 import { localeDetector } from "../../i18n/locales";
 import { PlayerTextures } from "../Entity/PlayerTextures";
 import { PUSHER_URL } from "../../Enum/EnvironmentVariable";
+import { CustomizeSceneName } from "./CustomizeScene";
```
moufmouf, line 12:
```suggestion
```

moufmouf:
Merged! Thanks @Waclaw-I !

PR #2020, opened 2022-03-29
Global voice-indicators

Waclaw-I:
Global voice-indicators (WIP)

- [x] Make use of proto well known types for changing player outline message
- [x] Send voice above / below treshold message
- [x] Display voice indicators above WOKA's head in reaction to PlayerDetailsUpdated message
- [x] Remove Group-specific code for voice indicators


moufmouf (approved):
:+1: I'm just wondering if we could get rid of the PlayerDetailsUpdatedMessageInterface interface and use the type created by ts-proto instead

PR #2026, opened 2022-03-31
Fix front build

fcecagno:


lukashass:
Hey @fcecagno, I just created #2025 which should fix your issue in a cleaner way. 😉 

As this is just a workaround for a bug in `vite-plugin-env-config` (and also might create problems with vite cleaning the dist folder on build), I'll close this in favor of #2025

PR #2044, opened 2022-04-06
Availability indicator

Waclaw-I:


moufmouf:
I love the small animation when switching from on state to the other :+1:

image

Regarding the design, could you try:

- to put the indicator a few pixels to the right (closer to the name, maybe 2 or 3 pixels only)
- to center the "indicator + name" block on top of the woka

PR #2159, opened 2022-05-03
change the way BiggestAvailableArea is calculated

Waclaw-I:
Currently works for:

- [x] embedded websites
- [x] videos


front/src/Stores/BiggestAvailableAreaStore.ts
```diff
@@ -1,112 +1,136 @@
...
-        if (htmlChildren.length === 0) {
-            return {
-                xStart: 0,
-                yStart: 0,
-                xEnd: game.offsetWidth,
-                yEnd: game.offsetHeight,
-            };
-        }
+    // map iframes to Box'es and remove ones outside of the game canvas space
+    const iframeBoxes: Box[] = Array.from(document.getElementsByTagName("iframe"))
```
moufmouf, line 11: Warning, some iframes should not be taken into account. We don't want to take into account the iframes that are directly embedded into maps.

front/src/Stores/BiggestAvailableAreaStore.ts
```diff
@@ -1,112 +1,136 @@
...
-                yEnd: game.offsetHeight,
-            };
-        } else {
-            return {
-                xStart: lastDiv.offsetLeft + lastDiv.offsetWidth,
-                yStart: lastDiv.offsetTop,
-                xEnd: game.offsetWidth,
-                yEnd: game.offsetHeight,
-            };
+    }
```
moufmouf, line 46: What if you were instead iterating on boxes to fill the occupiedSpace area?

Something like:

```
for (const box of iframeBoxes) {
    // The findBoxCoordinates returns the coordinates of the box in the occupiedSpace matrix
    [iStart, iEnd, jStart, jEnd] = findBoxCoordinates(box, xVertices, yVertices);
    for (i=iStart; i<iEnd; i++) {
        for (j=jStart; j<jEnd; j++) {
            occupiedSpace[i][j] = true;
        }

    }
}

**Review by moufmouf — COMMENTED** [review]
Woah, that's hard.

I added a few comments.
I think the first thing to do is probably to break the issue into small functions and to unit test each function

File: front/src/Stores/BiggestAvailableAreaStore.ts [hunk 3]
```diff
@@ -123,4 +97,39 @@ function createBiggestAvailableAreaStore() {
...
+        for (let x = left; x < xEnd; x += 1) {
+            // do not add squares with occupied parts
+            if (grid[y][x] === true) {
+                // no point in trying to find free square after this column
+                xEnd = x;
+                break;
+            }
+            squares.push({ xStart: left, yStart: top, xEnd: x, yEnd: y });
+        }
+    }
```
moufmouf, line 126:
```suggestion
    for (let y = top; y < grid.length; y += 1) {
        for (let x = left; x < xEnd; x += 1) {
            // do not add squares with occupied parts
            if (grid[y][x] === true) {
                // no point in trying to find free square after this column
                xEnd = x;
                break;
            }
        }
        squares.push({ xStart: left, yStart: top, xEnd: xEnd, yEnd: y });
    }
```

I think you can safely add a single square at the end of the inner for loop.
Indeed, the last square of the inner loop contains all the squares defined in the inner loop.

front/src/Stores/BiggestAvailableAreaStore.ts
```diff
@@ -1,112 +1,86 @@
...
+    const freeBoxes: Box[] = [];
+    for (let x = 0; x < occupiedGrid.length - 1; x += 1) {
+        for (let y = 0; y < occupiedGrid[x].length - 1; y += 1) {
+            freeBoxes.push(...findAllFreeBoxes(x, y, occupiedGrid));
         }
     }
+
+    const biggestFreeArea = freeBoxes
+        .map((box) => ({ box, area: getBoxArea(box, verticesX, verticesY) }))
+        .sort((a, b) => b.area - a.area)[0];
```
moufmouf, line 75: Micro optimization:

You can probably avoid the "sort" here and do a simple scan to find the biggest area.

The sort will be done in `O(n*log(n))` while simply scanning to find the biggest value can be done in `O(n)`

moufmouf (commented):
This definitely looks good to me. Good job!

I've just proposed 2 small optimizations.

moufmouf:
Hey @Waclaw-I ,

It's working great for the cowebsites, but not so much for the screen sharing.

See screenshot below:
screenshot

front/src/Stores/BiggestAvailableAreaStore.ts
```diff
@@ -1,112 +1,115 @@
...
-                xStart: 0,
-                yStart: lastDiv.offsetTop + lastDiv.offsetHeight,
-                xEnd: game.offsetWidth,
-                yEnd: game.offsetHeight,
-            };
+    const blockers: Box[] = [];
+
+    const screenOnTop = get(highlightedEmbedScreen)?.embed;
+    if (screenOnTop) {
+        //@ts-ignore
```
moufmouf, line 22: Could you remove the ts-ignore of this PR?

Here, the ".embed" property is defined as a `EmbedScreen` type which is:

http://github.com/thecodingmachine/workadventure/blob/6dd04e340c5f8896b72e2a6edb107b6f28943ea4/front/src/Stores/EmbedScreensStore.ts#L7-L16

```
export type EmbedScreen =
    | {
          type: "streamable";
          embed: Streamable;
      }
    | {
          type: "cowebsite";
          embed: CoWebsite;
      };
```

the iframe property exists only on `CoWebsite`, so you can remove the typescript warning by testing this:

```
if (screenOnTop.type === "cowebsite")
```

front/src/Stores/BiggestAvailableAreaStore.ts
```diff
@@ -33,21 +33,15 @@ function findBiggestAvailableArea(): Box {
...
-            }
+        const blockingElements = Array.from(document.getElementsByClassName("screen-blocker"));
+        for (const blocker of blockingElements) {
+            const bounds = blocker.getBoundingClientRect();
+            blockers.push({
+                xStart: bounds.x,
+                yStart: bounds.y,
+                xEnd: bounds.right,
+                yEnd: bounds.bottom,
+            });
```
moufmouf, line 44: Nice! I believe we can go even further.

Can we apply the "screen-blocker" class to the highlighted iframe too?
This way, we could also remove the `if (screenOnTop.type === "cowebsite")`, right?

moufmouf:
Hey Piotr, [3bf27eb](https://github.com/thecodingmachine/workadventure/pull/2159/commits/3bf27eb1049931b9d2697b842f7997a35297b3e8) does not seem to work. When I click the layout change button, repositioning does not occur

moufmouf:
Excellent! Thanks! :+1:

PR #2179, opened 2022-05-06
Update all docker nodejs container to the last LTS (16)

nolway:
Update all docker nodejs container to the last LTS (16)


moufmouf (changes requested):
Ok for the changes, but you also need to change the image of the Dockerfiles!
Otherwise, it will run in Node 16 in dev, but Node 14 in prod containers ;)

nolway:
> 



> Otherwise, it will run in Node 16 in dev, but Node 14 in prod containers ;)

Done =)

Issue #2171, opened 2022-05-06
Game crashes after camera configuration + silent zone


ValdoTR:
![image](https://user-images.githubusercontent.com/14997703/167096624-093b638d-c7a0-47e7-b520-7607cdf1d9ea.png)

If you click on "Edit you camera" in the menu and continue (blue button): then go into a silent zone and the game will crash.
![image](https://user-images.githubusercontent.com/14997703/167095954-aecb3050-1eb0-4d92-849e-5961dd9b2813.png)
(here I can't move but I see my video still rendered and moving)

PR #2185, opened 2022-05-09
unsubscribe from availabilityStatusStore

Waclaw-I:


ValdoTR:
@Waclaw-I I tested and the issue #2171  is resolved ;)

PR #2323, opened 2022-06-17
New chat service

gparant:
Create new application chat
 - [x] Take out chat of front and create dedicated chat service
 - [x] Use chat application ( chat.workadventure.localhost) in the front WorkAdvetnure
 - [x] Migrate user list design from front in chat
 - [x] Create new WebSocket connexion between the pusher and the chat


gparant:
Good job @cesar-cardinale, my first feedback: 

Capture d’écran 2022-07-01 à 21 00 44

PR #2370, opened 2022-07-03
Create new emoji sub menu

gparant:
- [x] Create sub menu,
 - [x] Update position emoji popup,
 - [x] Select and update emoji in the sub menu emoji by position,
 - [x] Save favorite emoji in navigator local storage,
 - [x] Define F1, F2, F3, F4, F5, F6 keybord to play emoji
 - [x] Update to remove icon for limited map menu 
 - [x] In mobile, delete number 1 - 2 - 3 - 4 - 5 - 6

https://user-images.githubusercontent.com/23724509/177036000-a77550fe-6a30-4ea2-b590-621da1a64609.mov


Lurkars:
I quite like! Just one suggestion, besides the edit, I would just add another picker button, so that you can select emoji directly for posting and not need to swap emoji to favorites every time (maybe you want a specific emoji one time for a specific moment, then it's complicated to change favorite first). 

Instead of button, best would be a drag'n'drop, so if you open picker, selected is posted and you can just d'n'd into favorites.

moufmouf:
I love it!

One suggestion:

F1-F6 keys are usually used for something else. For instance, on my browser, pressing F3 opens the search bar.

Maybe we could bind the icons to keys 1-6 instead?
(or CTRL-1 -> CTRL-6)?

ValdoTR:
I feel like seeing a new WorkAdventure! Awesome!

My suggestions (not exactly related to this PR but related to the UI design):
- Get rid of the "Press SPACE or touch here", WA can manage that itself depending on the user's device. We could also make this key customizable by the user.
- Add a lighter blue or grey color on top of the bar when the button is not active
- Invite button should have the same height as the action bar
- The WA menu should be before the action bar: WA -> actions -> Invite

BTW, very good idea imo to put the WA button here, that way users have more possibility when they add custom Button thanks to the scripting API.

Here is a prototype to explain my first point:
image
image

gparant:
> I love it!
> 
> One suggestion:
> 
> F1-F6 keys are usually used for something else. For instance, on my browser, pressing F3 opens the search bar.
> 
> Maybe we could bind the icons to keys 1-6 instead? (or CTRL-1 -> CTRL-6)?

@moufmouf is already bind to key 1, 2, 3, 4, 5, 6
=> https://github.com/thecodingmachine/workadventure/pull/2370/files#diff-836281ef693905ea7e26fee4fa81d75dab94a4c69935a94a1763cc5e78b2555fR170

moufmouf:
> @moufmouf is already bind to key 1, 2, 3, 4, 5, 6

Oh, ok. I need to press shift cause of the FR layout.
That would be awesome if we could have something binding to the key rather than the character but I believe it's almost impossible, so don't bother.

A possible improvement (for the future):

Now that we can spam the emoticon key, it would be super if several key presses (for instance on applause) was actually triggering a lot of emoticon going a bit higher in the air than a single one (just like Jitsi is doing)

PR #2423, opened 2022-07-15
WIP: Reworking Video UI

HimeShaman:


nolway:
I think the voice indicator isn't enough visible, can you try to change the color by the primary ?

There few issues in Firefox : 
- The left-bottom corner of the video container of others users isn't rounded : 
image
- Our video container isn't rounded:
image
- The loader and the mute icon aren't over the video and not enough visible, can you replace the color by the primary ?
image

cesar-cardinale:
I think there is a little bug :
You should hide the camera div and if the camera is enabled you show the div.
Right now it's the opposite.


https://user-images.githubusercontent.com/37698963/188461149-f9ec66d6-39ea-47d7-b251-03357e63f039.mov

cesar-cardinale:
> I think there is a little bug : You should hide the camera div and if the camera is enabled you show the div. Right now it's the opposite.
> 
>  Enregistrement.de.l.ecran.2022-09-05.a.15.29.22.mov

It's seems that this issue is still here ...


And also, I don't know if it's difficult but the right bar is not aligned to the action bar ... 🙄
Capture d’écran 2022-09-15 à 10 31 25

moufmouf:
The sound meter still feels a bit weird to me. It is not moving "enough" when a sound is detected. If the sound level stays the same, the sound meter does not move at all (but in a typical sound meter, the sound meter moves when there is a sound)

moufmouf:
I'm facing 2 weird behaviours:

image

After forming a bubble for the 3rd time, I'm seeing this.

The video does not stretch to the full container width and the wheel is still spinning.
Also, the videos in vertical mode  are not really on the top right corner. There is a huge margin.

moufmouf:
The sound meter looks much better though! Good job!

PR #2458, opened 2022-07-28
Chat sound

cesar-cardinale:
### Improvements
- [x] Emit sound when new message is received,
- [x] Disable sound when settings "Sounds of chat" is disabled in the menu,
- [x] Emit notification on desktop when new message is received,
- [x] Disable notification when settings "Notifications" is disabled in the menu,
- [x] Disable sound and notification when the chat is opened and the current thread is opened,
- [x] Write a lot of E2E tests to be sure that every feature is working well : 
  - [x] Chat fully loaded,
  - [x] Users list,
  - [x] Live zone,
  - [x] Send / receive message / file,
  - [x] Reaction,
  - [x] Answer to a message,
  - [x] Rank up / down and ban a user,
  - [x] Walk to

### Fix
- [x] Fix ejabberd.yml configuration for self-hosted solution,
- [x] Fix e2e test about the chat,
- [x] Fix same nickname in all mucRoom


front/chat/src/Media/MediaManager.ts
```diff
@@ -40,11 +46,13 @@ class MediaManager {
           break;
         case NotificationType.message:
           new Notification(
-            `${userName} ${get(LL).notification.message()}`,
+            `${userName} ${get(LL).notification.message()} ${forum !== null && get(LL).notification.forum() + ' ' + forum}`,
             options
           );
           break;
       }
+      console.log("Notification sent");
```
gparant, line 54:
```suggestion
```

front/chat/src/Xmpp/MucRoom.ts
```diff
@@ -753,6 +761,13 @@ export class MucRoom {
             } else {
               if (delay > this.lastMessageSeen) {
                 this.countMessagesToSee.update((last) => last + 1);
+                if (
+                  !get(chatVisibilityStore) ||
```
gparant, line 765: I don't understant the `chatVisibilityStore` usage?

We play sound and send notification only if the chat is open?

front/src/Api/IframeListener.ts
```diff
@@ -665,6 +677,35 @@ class IframeListener {
...
 
+    sendChatVisibilityToChatIframe(visibility: boolean) {
+        this.postMessageToChat({
+            type: "chatVisibility",
+            data: {
+                visibility,
+            },
+        } as IframeResponseEvent);
+    }
+    sendSettingsToChatIframe() {
```
gparant, line 688: The notification are sent only when the user change it. But I think that we need to sen them with user data?
Because if the user don't change settings and it's rather unsual, the notification settings will never defined in chat.

front/chat/src/Xmpp/MucRoom.ts
cesar-cardinale: It's the opposite,we play the sound and send the notification when :
- the chat is closed 
- or if the chat is open and that the current thread is not the thread where the new message comes from

front/src/Api/IframeListener.ts
cesar-cardinale: You're right, I send this settings when the user data are sent now.

PR #2548, opened 2022-08-29
Updates Cowebsite UI to match the new one

HimeShaman:


gparant:
I believe that we should use the close icon of the menu?

Screenshot 2022-09-05 at 17 13 30

PR #2619, opened 2022-09-18
Use same ejabberd image for dev as for prod

lukashass:
Fixes `Failed to start ejabberd application: Invalid value of option ca_file: Failed to read PEM file '/opt/ejabberd/conf/cacert.pem': no such file or directory`


lukashass:
Superseded by #2623

PR #2709, opened 2022-10-14
Refactor the whole chat architecture

cesar-cardinale:
Making the pusher out of the communication between the chat and Ejabberd.
Now the chat communicate directly via WebSocket to Ejabberd and it's better because the XMPP.js library contains some errors that make the pusher crash.
Delete all ideas of XmppClient in the pusher and migrate all the creation of the WS with Ejabberd in the chat.


cesar-cardinale:
close #2749

PR #2904, opened 2022-12-21
Audio & silent zone better UI

cesar-cardinale:
- [X] Change the UI of the audio range slider with the new design system.
- [X] Better silent zone indicator (taking less space).


play/src/front/Components/AudioManager/AudioManager.svelte
```diff
@@ -70,7 +70,9 @@
             let volume = HTMLAudioPlayer.volume;
             audioPlayerVol.value = "" + volume;
             audioPlayerVolumeIcon.classList.remove("muted");
-            if (volume < 0.3) {
+            if (volume == 0) {
+                onMute();
```
gparant, line 74: If the user defines 0, will be muted. But if he wants to up the volume, it's possible in muted position?

cesar-cardinale: Yes it is, it's only to show the Icon of sound in muted mode.

cesar-cardinale:
Also add in muted mode, the icon :
Capture d’écran 2022-12-22 à 11 01 05

PR #2935, opened 2023-01-03
Set the hair over the clothes

mlima95:


mlima95:
I set the hair over clothes, but if you choose a hoodie with hair, the result will be:

back hair over hood
left hair over hood
front over hoods
To avoid this, I recommend that you set your player's hair to bald, as shown here:

back bald hood
left bald hood
front bald hood


Perhaps in the future, we can set it to bald automatically when you wear a hood or set the hood as an accessory.

moufmouf:
Looks good to me.
:+1: on merging this.

@mlima95 You should bind this PR to the relevant issue.

@gparant Ok to merge this?

gparant:
@mlima95 go for merge 🚀

PR #2973, opened 2023-01-23
Hotfix: Fix emoji issue

mlima95:
ADD : 
- Replace twemoji with local emoji 
- Replace img tag with span tag 

GOAL : 
- Fix emoji issue

### DIFF
--- a/chat/index.html
+++ b/chat/index.html
@@ -37,6 +37,5 @@
 
         </div>
         <script type="module" src="/src/index.ts"></script>
-        <script async src="//cdn.embedly.com/widgets/platform.js" charset="UTF-8"></script>
     </body>
 </html>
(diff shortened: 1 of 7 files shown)


LindezaGrey:
just tested the 1.14.14 release.
for existing users the emojis look like this:
emoji fix
I had to delete the "emojiFavorite" key in the LocalStorage. Otherwise it works fine 👍

mlima95:
> just tested the 1.14.14 release. for existing users the emojis look like this: emoji fix I had to delete the "emojiFavorite" key in the LocalStorage. Otherwise it works fine 👍

Thank you very much for testing; we will fix this tonight.

moufmouf:
@LindezaGrey It should be fixed in v1.14.15. Emojis will be reset to their default value when migrating, but no "undefined" will appear anymore.

PR #3049, opened 2023-02-14
Add dynamic custom property for bindings areas

mlima95:
ADD : dynamic property for areas to the variable binding

### DIFF
--- a/libs/map-editor/src/GameMap/GameMapAreas.ts
+++ b/libs/map-editor/src/GameMap/GameMapAreas.ts
@@ -38,6 +38,13 @@ export class GameMapAreas {
             this.gameMap.tiledObjects
                 .filter((object) => ["zone", "area"].includes(object.class ?? ""))
                 .forEach((areaRaw: ITiledMapObject) => {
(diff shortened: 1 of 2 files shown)


libs/map-editor/src/GameMap/GameMapAreas.ts
```diff
@@ -38,6 +38,13 @@ export class GameMapAreas {
             this.gameMap.tiledObjects
                 .filter((object) => ["zone", "area"].includes(object.class ?? ""))
                 .forEach((areaRaw: ITiledMapObject) => {
+                    const properties = areaRaw.properties ?? [];
+
+                    for (const property of properties) {
+                        if (property.name === "dynamic" && property.type === "bool" && property.value) {
```
Waclaw-I, line 44: Sadly we cannot do it like that. Such chance would be like a backdoor for modifying Static Areas via ScriptingAPI without any safety measures.

Waclaw-I:
It will be taken care of the way it is described in this issue: https://github.com/thecodingmachine/workadventure/issues/3051

PR #3122, opened 2023-03-10
initial upper sorbian translation of chat window

zobadaniel:
Please add the upper sorbian translation for the chat window.


LindezaGrey:
i think the type should be "Translation" not "BaseTranslation"

zobadaniel:
Sry did not know that. I changed the type now.

PR #3204, opened 2023-04-21
[Temporary Don't Merge]Temporary DB

nolway:


.env.template
```diff
@@ -128,3 +128,6 @@ REPORT_ISSUES_URL=
 
 # LogRocket
 LOGROCKET_ID=
+
```
nolway, line 131: WIP to block the merge

PR #3261, opened 2023-05-25
Areas live streaming

cesar-cardinale:


moufmouf (commented):
I haven't had time to test this directly but here are a few comments on the code itself.

Woah, moving from uuid to id was quite a feat :)

PR #3563, opened 2023-10-16
add missing upper sorbian translations

zobadaniel:
Please include the missing translations of the map editor and the webrtc connection error into upper sorbian.


zobadaniel:
This PR now includes some more upper sorbian fixes, plus a fix for lower sorbian (changed error.ts) which triggered a tests failure. Please let me know if you need the languages in separate PRs.

PR #3741, opened 2024-02-26
[WS-1157] Apply management rules for new status 

ydiarra:
- Create modal to accept a bubble
- Create modal to go back to online status
- Apply management rules for new status


play/src/front/Components/ActionBar/AvailabilityStatus/Interfaces/AvailabilityStatusPropsInterface.ts
```diff
@@ -0,0 +1,20 @@
+import { AvailabilityStatus } from "../../../../../../../libs/messages";
```
gparant, line 1: use import "@workadventure/messages" because when we will compile that, it won't work

play/src/front/Components/ActionBar/AvailabilityStatus/Modals/ConfirmationModal.svelte
```diff
@@ -0,0 +1,33 @@
...
+<svelte:window on:keydown={onKeyDown} />
+
+<div
+    transition:fly={{ y: -900, duration: 500 }}
+    class="tw-flex tw-flex-col tw-text-white tw-bg-blue-900 tw-backdrop-blur tw-rounded-xl
+            tw-m-2 tw-pointer-events-auto
+            tw-bg-medium-purple/70 tw-max-w-sm tw-max-h-44 tw-self-center tw-justify-self-center "
+>
+    <section class="tw-flex tw-grow tw-items-center">
+        <slot>Define Description Part...</slot>
```
gparant, line 23: Translate the message

play/src/front/Components/ActionBar/AvailabilityStatus/Modals/NotificationPermissionModal.svelte
```diff
@@ -0,0 +1,35 @@
...
+    import { analyticsClient } from "../../../../Administration/AnalyticsClient";
+    import { localUserStore } from "../../../../Connection/LocalUserStore";
+    import { iframeListener } from "../../../../Api/IframeListener";
+    import ConfirmationModal from "./ConfirmationModal.svelte";
+
+    const confirmationModalProps: ConfirmationModalPropsInterface = {
+        handleAccept: () => {
+            notificationPermissionModalVisibility.close();
+            analyticsClient.settingNotification("true");
+            Notification.requestPermission()
```
gparant, line 14: It already exist in `MediaManager`, maybe use the same class or move this part in the `MediaManager` class

play/src/front/Phaser/Player/Player.ts
```diff
@@ -143,6 +144,15 @@ export class Player extends Character {
             x = x + 1;
         }
 
+        if (
+            activeEvents.get(UserInputEvent.MoveRight) ||
+            activeEvents.get(UserInputEvent.MoveLeft) ||
+            activeEvents.get(UserInputEvent.MoveDown) ||
+            activeEvents.get(UserInputEvent.MoveUp)
+        ) {
+            passStatusToOnlineWhenUserIsInSetableStatus();
```
gparant, line 153: MayBe you could use `handleKeyDownEvent` of the `UserInputManager` class in the `GameScene`. All events are used with this class or handler function created to this use case.

play/src/front/Components/ActionBar/AvailabilityStatus/Modals/ConfirmationModal.svelte
```diff
@@ -0,0 +1,33 @@
...
+    </section>
+    <footer class="tw-flex tw-justify-end tw-p-2 tw-backdrop-blur-2xl tw-rounded-b-2xl ">
+        <button class="tw-bg-blue-500 tw-rounded-2xl hover:tw-brightness-125" on:click={handleAccept}
```
gparant, line 26: To have the same design, use `light` class.
You can check another design created like "menu" and apply the same

play/src/front/Stores/MediaStore.ts
```diff
@@ -324,6 +324,10 @@ export const inJitsiStore = writable(false);
 export const inBbbStore = writable(false);
 export const isSpeakerStore = writable(false);
 
+export const doNotDisturbStore = writable(false);
+export const backInAMomentStore = writable(false);
+export const busyStore = writable(false);
+
```
moufmouf, line 330: Instead of these 3 stores (that could be in conflict), I wonder if we should not have a single store called `requestedStatusStore` that contains the store the user wants.

Note it can be different from the status the user is granted in the end (because it can be overridden by a silent zone, etc...)

moufmouf:
image


A click out of the status selector (anywhere) should close the status selector. Right now, we need to click on the status button to close it.

moufmouf:
I think we should change the color (or background color) of the selected option.

image

Right now, the selected option is grayed out. It carries the idea it cannot be selected (this is true) but not that this the active status.

We could take a look at how the others do it. For instance Google:
image

The tick makes it clear what status is selected.

PR #3766, opened 2024-03-05
Improve UI speaker zone

gparant:


gparant:
Review for @ydiarra 

When a User is in the meeting, a rule implemented is that we cannot change the status and it works 💪

I think that we need to show information at the user throughout the tooltip to show what the status can't be changed during a meeting. Examples of fun interaction should be "Have a good meeting 🤓".

Screenshot 2024-03-06 at 08 27 09

PR #3830, opened 2024-03-28
Visit card personal on entering area

matthintosh:
On entering in a personal area, a user will see the visit card of the area owner.


libs/map-editor/src/types.ts
```diff
@@ -131,7 +131,7 @@ export const PersonalAreaPropertyData = PropertyBase.extend({
     type: z.literal("personalAreaPropertyData"),
     accessClaimMode: PersonalAreaAccessClaimMode,
     allowedTags: z.array(z.string()).default([]),
-    ownerId: z.string().default(""),
+    ownerId: z.string().nullable(),
```
gparant, line 134: use optional, by default the proto message serialiser can't take null option

libs/messages/src/JsonMessages/MemberData.ts
```diff
@@ -4,6 +4,7 @@ export const MemberData = z.object({
   id: z.string(),
   name: z.string().nullable(),
   email: z.string().nullable(),
+  visitCardUrl: z.string().nullable(),
```
gparant, line 7: same

play/src/pusher/services/AdminApi.ts
```diff
@@ -1022,6 +1022,34 @@ class AdminApi implements AdminInterface {
         });
         return response.data ? response.data : [];
     }
+
+    /**
+     * @openapi
+     * /members:
```
gparant, line 1028: that is correct ?
If we search by `memberUUID`, we could use "/members/{uuid}"?

matthintosh: I added the missing parameter for openApi documentation.

PR #3868, opened 2024-04-17
Fix remove entity personal area changes

matthintosh:
A user see a pop-up to ask him if he wants to keep or delete entities in personal area while removing it, revoking the owner or removing personal area property.


gparant:
Something is broken


https://github.com/workadventure/workadventure/assets/23724509/831df5f8-55cf-4d44-9ee9-8f22c96eab6b

Issue #4067, opened 2024-07-17
[MAP EDITOR] Map freezes when entering an image tag


ValdoTR:
I tried uploading a custom image and giving it a tag name.
The map freezes immediatly after attempting to write something in the tag input:
![image](https://github.com/user-attachments/assets/4579dda4-6883-446c-980a-39ab732aa64b)

ValdoTR:
It seems to be related to uppercase letters.

PR #4069, opened 2024-07-17
Input tag issue with upper case

matthintosh:
Fix #4097


ValdoTR:
Hi, the issue #4067 has not been closed automatically after this one has. Weird.

PR #4253, opened 2024-09-19
draft: [MATRIX] Merged message

gparant:
When user sent multiple message in the same minute, we merge with the previous

### DIFF
--- a/play/src/front/Chat/Components/Room/Message.svelte
+++ b/play/src/front/Chat/Components/Room/Message.svelte
@@ -19,6 +19,7 @@
 
     export let message: ChatMessage;
     export let reactions: MapStore<string, ChatMessageReaction> | undefined;
+    export let isMerged = false;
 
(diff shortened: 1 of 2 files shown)


moufmouf (commented):
Pretty good.

One things that affrays me a little is performance. Each time we receive an event (message, reaction, etc...), we are going to recompute the complete `groupedMessagesStore`.

That might end up being costly (it is just a guess).

I'm wondering if we could not do this processing at the moment we iterate the message list (instead of running it through a new store).

I'm not 100% sure how to do this, but it might look like this:

```
{#each $messages as message, key (message.id)}
    {@const previousMessage = $messages[key - 1] }
    {#if message.type = "..." && message.sender.uuid === previousMessage.sender.uuid && ...}
      <Message isMerged={true}>
    {/else}
       <Message isMerged={false}>
    {/if}
```

Do you think that could work? It would probably save recomputing everything each time a new message comes in.

PR #4260, opened 2024-09-20
Stabilize e2e part.2

ydiarra:


ydiarra:
image

moufmouf:
I understood the issue in Firefox!

When we mute the audio, for some reason, the complete video container div is refreshed.
As a consequence, it is closing the action bar. Therefore, you cannot click on video mute button anymore. We would need to click on the "open menu" item again.

moufmouf:
We still have one flaky test (absolutely each time):

image

I think this is because on the second try, the key already exists so something fails always the first time and is successful always on the second try.

PR #4374, opened 2024-11-05
Arabic Translation by klinshy

ydiarra:


klinshy:
Hi @ydiarra, Ive had the chance to look on the staging env with @BAWES

According to him, the translations are mostly allright. There are few words for example in the Audio settings where "unlimited" says "undefined", or "recommended"-AudioLevel word doesnt make sense.

Context opened in html popups in the menu, mix words as we were already been afraid of. That means that english words may end up at the end or beginning of the sentence, where they shouldnt be.
This message at the chat does illustrate it. Khalid first sended ONE  chat, and words got mixed up wrong. Then he sended each part of the message in the order as it should hat been created.
image

As you've already tested @ydiarra, the `dir="rtl"` html object is the way to go, to have everything smoothed out. 
In the following recording, Khalid does test and add the `dir="rtl"` to individual sections to achieve the desired order and display. What he does different from your initial test is that he doesnt  put `dir="rtl"`, in the global head of the page.
From what we all saw in the staging, the UI has been fully moved to the opposite sides. 
[Watch Screenrecord](https://rec.cocreation.world/KDtRWLBv)


Im drawing the conlusion that the best way to go (eventually) is to have somehow in general `dir="rtl"` present globally, without actually mirroring locations of the Buttons, Chatboxes etc. to the other side. **But only their content**


The final conclusion now is the following:
Our tests show Workadventures versatility well, its a great MVP to show what could be achieved. For a final and good-looking public release, it would require one further but minor tweak, which is to add `dir="rtl"` for content only, and not for the menus and buttons themselves.

@moufmouf @gparant @ValdoTR (pinging you guys here as I believe you might be interested, too)
Side note: Khalid pointed out that it is important for members of his Arab cultural group to use a platform ‘super-easy’. It must be quickly and easily clear HOW the platform is used. If you first have to find and look up a documentation, you quickly lose interest.
I also share his observation from my own experiences in my circle and can report back that some tooltips would be helpful. For first-time users, important knowledge must be incorporated into the processes.


Thank you everyone!

PR #4448, opened 2024-12-04
Adding main fixs to chat

HugoAverty:
- Removing dots Icon in header (useless)
- Increasing message size for proximity and chats rooms
- Changing avatar rounded


HugoAverty:
Fixing height of user list

PR #4463, opened 2024-12-16
fix special characters

ydiarra:


moufmouf (changes requested):
The underlying problem is that we should send a "formatted_body" along the "body" in the message.
Which means Markdown transformations should happen when sending a message (and not when reading it).

PR #4484, opened 2024-12-30
Limit member search to world members and Matrix-known users

ydiarra:
TODO : 

- [x] Allow user to enter any Matrix address

### DIFF
--- a/play/src/front/Chat/Components/SelectMatrixUser.svelte
+++ b/play/src/front/Chat/Components/SelectMatrixUser.svelte
@@ -4,12 +4,12 @@
     import Select from "svelte-select";
     import LL from "../../../i18n/i18n-svelte";
     import { gameManager } from "../../Phaser/Game/GameManager";
-    import { searchChatMembersRule } from "./Room/searchChatMembersRule";
-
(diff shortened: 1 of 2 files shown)


play/src/front/Chat/Components/SelectMatrixUser.svelte
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        return validItems;
+    }
+
+    async function loadItems() {
+        try {
+            return await searchWorldMembers(filterText);
+        } catch (error) {
+            console.error(error);
```
gparant, line 58: If there is an error when we try to search world member, could we display this error?

play/src/front/Chat/Components/SelectMatrixUser.svelte
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    onMount(() => {
+        loadItems()
+            .then((newItems) => {
+                items = newItems;
+            })
+            .catch((error) => {
+                console.error(error);
```
gparant, line 70: If there is an error when we try to load world member, could we display this error?

PR #4511, opened 2025-01-17
Faster test

Powopaf:


Powopaf:
All tests in chromium pass except for second test in variable.spec.ts (docker) and second test in meeting.spec.ts. 
Test in meeting is unstable

Powopaf:
all chromium test are valid same in firefox expect for one

PR #4516, opened 2025-01-23
fix webkit test refonte

Powopaf:
branch create to fix test run on WebKit. I also switch to playwright v-1.50


Powopaf:
webkit test don't run in CI but do work when test locally

PR #4549, opened 2025-02-12
Detect & propose to use new device

gparant:
Detect new device and propose to use it

Fix
When the new device appears and is set as the preferred one, utilize it. Do not let the navigator to use default device if exist.

https://github.com/user-attachments/assets/597bbcda-27f2-4dc6-a21c-2c6a4d0e109c


moufmouf (commented):
Super nice!

When we merge to the new design, we will probably move the message to a popup at the bottom of the screen (Hugo thinks we should do this for all messages that are not directly related to the Woka which I tend to agree).
No need to do it here though, we can move this during the merge.

PR #4583, opened 2025-03-08
Fix contrib/docker/docker-compose.prod.yaml

pizkaz:
Adds missing router->service mappings for play container.


pizkaz:
The Workadventure setup we use is behind a reverse-proxy (nginx) which handles TLS, so I ran into the problem right away. ^^

PR #4715, opened 2025-04-08
Fix tooltip by using floating ui

Maison-hub:
image


play/src/front/Components/MapEditor/PropertyEditor/AddPropertyButton.svelte
```diff
@@ -8,11 +10,31 @@
     export let disabled = false;
     export let testId = undefined;
     const dispatch = createEventDispatcher();
+
+    if (!testId) {
```
gparant, line 14: I think it could be better to use hook of Svelte like "onMounted" to change the value

gparant: Or force to have testId for any button created

PR #4768, opened 2025-04-22
Add new analytics event for posthog

Maison-hub:


play/src/front/Chat/Components/Room/CreateFolderModal.svelte
```diff
@@ -35,6 +36,7 @@
         } finally {
             loadingFolderCreation = false;
         }
+        analyticsClient.createMatrixFolder();
```
gparant, line 39: May you could move it in the finally ?

play/src/front/Administration/AnalyticsClient.ts
```diff
@@ -811,5 +779,122 @@ class AnalyticsClient {
...
+            .catch((e) => console.error(e));
+    }
+    startMatrixEncryptionConfiguration(): void {
+        this.posthogPromise
+            ?.then((posthog) => {
+                posthog.capture("wa_start_matrix_encryption_configuration");
+            })
+            .catch((e) => console.error(e));
+    }
+    externalModuleChatBandClick(externalModuleName: string, action: string): void {
```
gparant, line 882: Is created but changes not implemented

gparant (changes requested):

PR #4779, opened 2025-04-24
Fix todos and calendar popups

Maison-hub:
# Now ✨
image
image

# Before 🙈
image
image


play/src/front/Components/TodoList/TodoList.svelte
```diff
@@ -25,43 +26,70 @@
...
+            <div class="header-container pb-4">
+                <div class="flex flex-row items-start justify-between">
+                    <h3 class="text-lg text-left py-2">
+                        <img
+                            draggable="false"
+                            src={todoListPng}
+                            class="w-6 -mb-1 mx-2"
+                            alt={$LL.menu.icon.open.todoList()}
+                        />
+                        To Do 📋 (beta)
```
gparant, line 45: Update "beta"

PR #4805, opened 2025-04-28
Dont handle shortcut when typing in the chat

Maison-hub:
This pull request introduces a safeguard in the `ChatSidebar.svelte` file to prevent unintended key events when certain input elements are focused.

Key changes:

* [`play/src/front/Chat/ChatSidebar.svelte`](diffhunk://#diff-4682a84869fc4af44dcb179f9af8319a7b39d29b1333d320c4192709ceb66ee5R28-R35): Added a check in the `onKeyDown` function to ignore key events when the focused element is an `INPUT`, `TEXTAREA`, or has the `data-testid` attribute set to "messageInput". This prevents interference with user inputs in these fields.


play/src/front/Chat/ChatSidebar.svelte
```diff
@@ -25,6 +25,14 @@
     const isUserListEnabled = gameScene.room.isChatOnlineListEnabled;
 
     function onKeyDown(e: KeyboardEvent) {
+        const activeElement = document.activeElement as HTMLElement | null;
+        const dataTestId: string | null | undefined = activeElement?.getAttribute("data-testid");
```
gparant, line 29: `data-testid` is used especially to playright testing. `activeElement.tagName === "INPUT"` must be sufficient?

gparant (changes requested):

PR #4855, opened 2025-05-21
Fixed the cowebsite button's UI problem

jadekeina:


moufmouf:
I think we can aim for a better solution.

Current solution:
image

It's a bit counter intuitive compared to what users are used to.
The idea of those tabs is to map as close as possible to real tabs of a web browser.

image

image

image

Behaviour of chrome when shrinking: the title of the tab is hidden first (the icon stays displayed)
After a breakpoint, the close button is hidden, except for the active tab
When it gets really tiny, only the icon is displayed (or for the active tab, only the cross)

We could attempt the same thing.

moufmouf (changes requested):
See https://github.com/workadventure/workadventure/pull/4855#issuecomment-2903605644

PR #4897, opened 2025-06-06
Fix area info duration

Maison-hub:
Duration settings was not save in tooltip area


play/src/front/Components/MapEditor/PropertyEditor/TooltipPropertyButton.svelte
```diff
@@ -17,6 +17,7 @@
     }>();
 
     function onValueChange() {
+        console.log("onValueChange", property.content, property.duration);
```
gparant: A supprimer de la MR

PR #4971, opened 2025-07-18
Drag drop files

jozzzzzz:


play/src/front/Components/GameOverlay.svelte
```diff
@@ -54,6 +56,11 @@
         <LoaderScene />
     </div>
 {/if}
+{#if $draggingFile}
```
gparant, line 59: Who can drop file, I think that we could manage right here?

You can help you with the rules used for the map editor.

play/src/front/Components/PopUp/PopupDropFileEntity.svelte
```diff
@@ -0,0 +1,136 @@
...
+            id: uuidv4(),
+            file: fileAsUint8Array,
+            name: file.name,
+            propertyId,
+        };
+
+        const roomConnection = scene.connection;
+        if (!roomConnection) throw new Error("No room connection");
+
+        new UploadFileFrontCommand(fileToUpload).emitEvent(roomConnection);
```
gparant: how do you manage the UI/UX if the upload is not working?

play/src/front/Components/PopUp/PopupDropFileEntity.svelte
```diff
@@ -0,0 +1,136 @@
...
+            width: 50,
+        };
+
+        mapEditorCopiedEntityDataPropertiesStore.update((properties) => {
+            const newProperties = properties ? [...properties] : [];
+            newProperties.push(property);
+            return newProperties;
+        });
+
+        analyticsClient.toggleMapEditor(true);
```
gparant: We could change that and use new analitic to the upload drag and drop feature 🙏

play/src/front/Phaser/FileUpload/FileListener.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    public dropHandler(event: DragEvent) {
+        event.preventDefault();
+        event.stopPropagation();
+
+        const userIsAdmin = gameManager.getCurrentGameScene().connection?.isAdmin();
+        const userIsEditor = gameManager.getCurrentGameScene().connection?.hasTag("editor");
+
+        if (!userIsAdmin && !userIsEditor) {
+            warningMessageStore.addWarningMessage("You do not have the rights to upload files on this map", {
```
jozzzzzz: add button to connect when disconnected

jozzzzzz: and add errors in i18n

play/src/front/Phaser/FileUpload/FileListener.ts
```diff
@@ -0,0 +1,124 @@
...
+        event.stopPropagation();
+
+        const userIsAdmin = gameManager.getCurrentGameScene().connection?.isAdmin();
+        const userIsEditor = gameManager.getCurrentGameScene().connection?.hasTag("editor");
+
+        if (!get(userIsConnected)) {
+            popupStore.addPopup(
+                PopUpConnect,
+                {
+                    message: get(LL).mapEditor.entityEditor.errors.dragNotConnected(),
```
gparant: Why send the message via the props and not directly use directly "dragNotConnected" inside "popupConnect" component?

jozzzzzz: So we can use it somewhere else if needed

PR #4997, opened 2025-07-31
remove duplicate buttons on woka menu

jozzzzzz:


play/src/front/Components/ActionsMenu/WokaMenu.svelte
```diff
@@ -42,6 +42,18 @@
                     return 0;
                 }
             });
+
+            // Remove actions with same actionName
+            const uniqueActions: string[] = [];
+
+            sortedActions = sortedActions.filter((action) => {
```
gparant, line 49: It's not here that we need to remove action added, it's in the store. I will close this merge request because the fix was apply in new MR: #5000

Thanks 💪

PR #4999, opened 2025-08-01
limit entities placement in area

jozzzzzz:


libs/map-editor/src/GameMap/GameMapAreas.ts
```diff
@@ -102,15 +102,41 @@ export class GameMapAreas {
...
+            y: entityCenterCoordinates.y - height / 2,
+        };
+        const bottomRightCoordinates = {
+            x: entityCenterCoordinates.x + width / 2,
+            y: entityCenterCoordinates.y + height / 2,
+        };
+
+        let validAreas: AreaData[] = areas;
+
+        if (floating) {
```
gparant: If that that could be the same for no floating entity?

PR #5010, opened 2025-08-06
Fixed the UI problems of the Space Error

jadekeina:


jadekeina:
Capture d’écran 2025-08-07 à 16 05 53

play/src/front/Space/Utils/UserLookup.ts
moufmouf, line 1: Instead of merging this PR, we will make sure lookupUserById is completely removed (in the Livekit version, users should be part of the space BEFORE a video is started, so this category of error should not happen anymore)

PR #5015, opened 2025-08-09
:recycle: Early exit in isUserHasWriteAccessOnAreaForEntityCoordinates

homersimpsons:
Notes:
- `npm run lint` in `libs/map-editor` raises 33 errors, maybe a CI issue
- I find that `overlapping` is a bit misleading in the name, because this checks whether a point is within a rectangle bounds (assuming horizonta and vertical sides), maybe this should be named `isPointInRectangle`
- `floating` parameter is unused
- Maybe we should not use `getAreasOnPosition` to have a dedicated `getAreasOnRectangle` 
- (not sure about the math) but as we assume rectangle with horizontal and vertical sides it is probably possible to remove the check on the center to just check for both corners, with someting along the following. Should I update the logic to avoid the extra check?
    ```typescript
        return this.getAreasOnPosition(topLeftCoordinates).some(
            (area) =>
                MathUtils.isOverlappingWithRectangle(bottomRightCoordinates, area) &&
                (this.isUserHasWriteAccessOnAreaByUserTags(area, userConnectedTags) || this.isAreaOwner(area, userUUID))
        );
    ```
- `getAreasOnPosition` could use `.filter` instead of `for` + `push`

Kind of a follow-up of #4999

### DIFF
--- a/libs/map-editor/src/GameMap/GameMapAreas.ts
+++ b/libs/map-editor/src/GameMap/GameMapAreas.ts
@@ -100,7 +100,6 @@ export class GameMapAreas {
         height: number,
         floating: boolean
     ): boolean {
-        const areas = this.getAreasOnPosition(entityCenterCoordinates);
         const topLeftCoordinates = {
             x: entityCenterCoordinates.x - width / 2,
             y: entityCenterCoordinates.y - height / 2,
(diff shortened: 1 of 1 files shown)


homersimpsons:
It seems that all playwright tests are failing with the same error:
```log
Run docker compose -f docker-compose.yaml -f docker-compose-oidc.yaml -f docker-compose.e2e.yml logs > tests/playwright-report/docker-compose.log
/home/runner/work/_temp/<uuid>.sh: line 1: tests/playwright-report/docker-compose.log: No such file or directory
```

I have no idea how to fix this as my change seemingly does not impact those tests.

homersimpsons:
I rebased this on `develop`, but it looks like `develop` is a bit outdated (3 commits ahead, 331 commits behind `master`) hence maybe I should target `master`, @moufmouf ?

homersimpsons:
The failure is just timeouts. I cannot re-run the job. I believe the tests are flaky.

moufmouf:
Indeed, you definitely should target master. We use develoo when we have a big feature coming, but right now, we are more in a patch phase, so we directly work on master 👍

homersimpsons:
> you definitely should target master

Done. CI is failing (https://github.com/workadventure/workadventure/actions/runs/21884255121/job/63175437894?pr=5015) on an unrelated check I think and I cannot re-launch the jobs.

PR #5061, opened 2025-08-27
add file size validation in front for drag & drop

ydiarra:


moufmouf (commented):
We shouldn't do that on drag'n'drop only.
We should also do that for the input fields in teh map editor.

PR #5385, opened 2025-11-24
Add individual volume controls for each stream

ydiarra:


moufmouf:
Nice!

A few improvements (not related directly to this PR but in the same scope):

1. We should change the labels of the "Mute audio" / "Mute video" buttons depending on the user.

If the user is NOT an admin, we should say "Ask to mute audio" instead, and when we click the button, a notification should pop up saying "A notification was sent to XXX to mute their microphone".

image

2. The "Mute audio" / "Mute video" in the context of a screensharing should be fixed. Today, they mute the user, not the screenshare.

PR #5394, opened 2025-11-27
Wait 20sec before switch back to webrtc

ydiarra:


back/src/Model/CommunicationManager.ts
```diff
@@ -84,9 +188,10 @@ export class CommunicationManager implements ICommunicationManager {
             if (this._toFinalizeState) {
                 try {
                     this.finalizeState(this._toFinalizeState);
-                } catch (e) {
-                    console.error("Error while finalizing state:", e);
-                    Sentry.captureException(e);
+                } catch (error) {
+                    const err = error instanceof Error ? error : new Error(String(error));
+                    console.error("Error while finalizing state:", err);
+                    Sentry.captureException(err);
```
moufmouf: Probably AI slop. You can keep the previous version, it's perfect.

back/src/Model/CommunicationManager.ts
```diff
@@ -6,57 +6,161 @@ import { WebRTCState } from "./States/WebRTCState";
...
+        // Check if we should stay in LiveKit (number of users > MAX_USERS_FOR_WEBRTC)
+        // If totalUsers <= MAX_USERS_FOR_WEBRTC, we do NOT cancel (switch will proceed)
+        const totalUsers = this.space.getAllUsers().length;
+        const shouldStayInLivekit = totalUsers > Number(MAX_USERS_FOR_WEBRTC);
+
+        if (shouldStayInLivekit) {
+            this.cancelPendingStateSwitch();
+        }
+        // If shouldStayInLivekit is false (totalUsers <= MAX_USERS_FOR_WEBRTC),
+        // we do nothing and let the switch proceed after the delay
```
moufmouf: Something is bothering me with this implementation.

So far, it was the responsibility of the `XXXState` classes to know if they should switch and what the next state was. This is implemented in `XXXState.shouldSwitchToNextState` methods.

Now, the `CommunicationManager` is also aware of business rules because it is starting looking at user counts. I'm afraid this might be hard to maintain in the long run (too many business rules in too many places).

We should probably either:

1. Centralize all business rules in one class (maybe CommunicationManager), and CommunicationManager would be in charge of instantiating all states and managing transitions
2. Or if we want to keep it the way it is, we should find a generic way to say if we should "still" switch or not (maybe by making the XXXState.shouldSwitchToNextState methods public?)

Also, the XXXState.handleUserAdded could return something like:

```
XXXState.handleUserAdded: {
   nextStatePromise: Promise<XXXState>|undefined,
   abortNextStateController: AbortController,
}
```

This way, the 20 seconds timeout could be handled by the `LivekitState` itself (and the CommunicationManager can abort it using the abortController passed in parameter)

PR #5402, opened 2025-12-01
Fix Matrix Chat "Connecting to server" Infinite Loading Issue

BAWES:
@moufmouf this pull request shows the changes made to resolve the issue. Not 100% sure if it's the correct way to go about it and will need testing.

## Problem

After successful Matrix login, users see "Connecting to server..." indefinitely, and no Matrix API requests are made.

## Root Causes

- Timing issue: `getChatConnection()` was called before userIsConnected was set to `true`
Location: `play/src/front/Phaser/Game/GameManager.ts:258`
The check `if (matrixServerUrl && get(userIsConnected))` failed because `userIsConnected` was still `false` when called early in initialization
Result: Matrix client was never created, returning `VoidChatConnection` instead
- Blocking clearStores() call
Location: `play/src/front/Chat/Connection/Matrix/MatrixClientWrapper.ts:180`
When Matrix user ID changed, `await this.client.clearStores()` could hang indefinitely on IndexedDB operations
Result: Client creation never completed
- Blocking `initRustCrypto()` call
Location: `play/src/front/Chat/Connection/Matrix/MatrixChatConnection.ts:290`
`await this.client.initRustCrypto()` could hang, blocking `startClient()` and preventing sync from starting
Result: No Matrix API requests were made

## Fixes Implemented

- Wait for user connection with timeout
`play/src/front/Phaser/Game/GameManager.ts`
Check `localUserStore.isLogged()` directly (more reliable than store)
If not connected, wait up to 10 seconds for userIsConnected to become true
Proceed with whichever becomes true first
- Made `clearStores()` non-blocking
`play/src/front/Chat/Connection/Matrix/MatrixClientWrapper.ts`
Run `clearStores()` in the background (fire-and-forget)
Return client immediately without waiting for store clearing
- Made `initRustCrypto()` non-blocking
`play/src/front/Chat/Connection/Matrix/MatrixChatConnection.ts`
Initialize crypto in the background with void (async () => {...})()
Start client sync immediately without waiting for crypto
Crypto initializes in parallel; if it fails, client continues

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -61,6 +61,12 @@ OPENID_SCOPE=
 OPENID_USERNAME_CLAIM=
 OPENID_TAGS_CLAIM=
 
+#Chat
+MATRIX_API_URI=
+MATRIX_PUBLIC_URI=
+MATRIX_ADMIN_USER=
+MATRIX_ADMIN_PASSWORD=
+MATRIX_DOMAIN=
 
 # Whether the user can choose its name or if the name is dictated by OpenID.
 # Can be one of "user_input", "allow_override_opid", "force_opid"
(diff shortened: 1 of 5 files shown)


BAWES:
this has been resolved in a recent commit and is no longer required

PR #5425, opened 2025-12-05
feat: add guest access

ThomasPerez91:


libs/eslint-config/index.js
moufmouf, line 1: You should probably not modify this file since it is not related to the current PR.

play/src/front/Phaser/Game/GameManager.ts
```diff
@@ -80,6 +84,37 @@ export class GameManager {
         console.info("Preferred audio input device: " + preferredAudioInputDeviceId);
         console.info("Preferred video input device: " + preferredVideoInputDeviceId);
 
+        // --- MODE INVITÉ ---------------------------------------------
+        if (this.isGuestMode()) {
+            if (!this.playerName) {
+                const random = Math.floor(Math.random() * 10000)
+                    .toString()
+                    .padStart(4, "0");
+                this.playerName = `Invité-${random}`;
```
moufmouf: "Invité" is French. I don't think we should translate it using the i18n package (because we want this to be the same for all users, whatever their language is). I think we should add an environment variable: `DEFAULT_GUEST_NAME`, and the default value for this env variable should be "".

If `DEFAULT_GUEST_NAME` is set, we can assume we are in "Guest mode", so you can probably remove the `GUEST_MODE` env var.

play/src/front/Phaser/Game/GameManager.ts
```diff
@@ -80,6 +84,37 @@ export class GameManager {
         console.info("Preferred audio input device: " + preferredAudioInputDeviceId);
         console.info("Preferred video input device: " + preferredVideoInputDeviceId);
 
+        // --- MODE INVITÉ ---------------------------------------------
+        if (this.isGuestMode()) {
+            if (!this.playerName) {
+                const random = Math.floor(Math.random() * 10000)
+                    .toString()
+                    .padStart(4, "0");
```
moufmouf: I'm not sure everybody will want a random identifier after the guest name. Some users might just want anyone to be called "guest" without anything after. We could add an additional env var to configure that: `GUEST_NAME_APPEND_RANDOM_NUMBERS`.

play/src/front/Phaser/Game/GameManager.ts
```diff
@@ -80,6 +84,37 @@ export class GameManager {
...
+                this.playerName = `Invité-${random}`;
+                localUserStore.setName(this.playerName);
+            }
+
+            if (!this.characterTextureIds || this.characterTextureIds.length === 0) {
+                const defaultGuestTextureId = "male4";
```
moufmouf: Same thing here, some users will want to customize the default texture. You can add an environment variable for that: `DEFAULT_GUEST_TEXTURE`.

moufmouf (changes requested):
Thanks a lot!

Below some comments.

libs/eslint-config/index.js
ThomasPerez91: In this branch, Husky fails during the ESLint pre-commit check.
When ESLint runs with --fix, I get the following error on every modified file:

Parsing error: Enabling "project" does nothing when "projectService" is enabled. You can remove the "project" setting

Example:
`/src/common/FrontConfigurationInterface.ts:0:0
  error  Parsing error: Enabling "project" does nothing when "projectService" is enabled
`

This happens for every file touched in the commit. It looks like ESLint refuses to run when both project and projectService are configured at the same time.

The only way I found to make the commit pass is temporarily commenting the project line inside libs/eslint-config/index.js, like this:

`parserOptions: {
  projectService: true,
  tsconfigRootDir: import.meta.dirname,
  // project: "./tsconfig.json",
  extraFileExtensions: [".svelte"],
}`

With this change, Husky/ESLint runs successfully and the commit can be validated.

PR #5427, opened 2025-12-05
feat(background): Add WebGL shader for confidence mask filtering

moufmouf:
Add a custom MaskFilterShader class that applies threshold-based filtering to MediaPipe's segmentation confidence masks before compositing.

The shader implements the following filter logic:
- Values below lowThreshold (0.3) are clamped to 0 (background)
- Values above highThreshold (0.9) are clamped to 1 (foreground/person)
- Values in between are smoothly interpolated using GLSL smoothstep

Key features:
- Full WebGL2 implementation with vertex and fragment shaders
- Supports both Float32Array mask data and WebGL texture inputs
- Handles Y-axis flipping between MediaPipe mask and canvas coordinates
- Includes fallback for systems without EXT_color_buffer_float extension
- Proper WebGL state management (saves/restores previous state)
- CPU fallback method (filterMaskCPU) for non-WebGL scenarios

This replaces the direct use of DrawingUtils.drawConfidenceMask() with our custom shader, giving us control over the mask filtering parameters to improve segmentation edge quality.

The thresholds can be dynamically adjusted via updateConfig() method.

### DIFF
--- a/play/src/front/Components/PopUp/PopUpBackgroundCamera.svelte
+++ b/play/src/front/Components/PopUp/PopUpBackgroundCamera.svelte
@@ -35,6 +35,14 @@
         backgroundConfigStore.setBackgroundVideo(videoUrl);
     }
 
+    function setLowThreshold(value: number) {
+        backgroundConfigStore.setLowThreshold(value);
+    }
+
+    function setHighThreshold(value: number) {
+        backgroundConfigStore.setHighThreshold(value);
+    }
+
     function resetBackground() {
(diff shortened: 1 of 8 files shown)


moufmouf:
Closing this PR:

The WebGL shader in this PR is working as expected.
But after adding the WebGL shader to smooth composition and trying to configure it for optimal performance, the conclusion is that the best results are when the shader is disabled. So no need to add additional complexity, we can keep doing a simple composition without advanced smoothing.

PR #5452, opened 2025-12-10
Add new css to improve UX for resize handler

gparant:


moufmouf:
Not a big fan of this bump animation. The first time I saw it, I thought it was a graphical glitch. But there is maybe some idea here. We could try to grow slightly the handle bar when hovered?

Also, on a side note, Tailwind has built in animations: https://tailwindcss.com/docs/animation#adding-a-bounce-animation

PR #5471, opened 2025-12-16
feat: locate user on the map

gparant:


play/src/front/Chat/Components/UserList/User.svelte
```diff
@@ -60,6 +60,24 @@
...
+        // Get remote player data to extract userId
+        const remotePlayerData = gameScene.getRemotePlayersRepository().getPlayerByUuid(user.uuid);
+        if(remotePlayerData == undefined) return;
+
+        // Get the actual RemotePlayer sprite from MapPlayersByKey using userId
+        const remotePlayer = gameScene.MapPlayersByKey.get(remotePlayerData.userId);
+        if(remotePlayer == undefined) return;
+
+        // Activate the remote player (opens woka menu)
+        remotePlayer.activate();
```
moufmouf: This will fail if the user is far away (not part of the viewport or nearby)

If the user is far away, the pusher will optimize by not sending the position of the user. Therefore, the user will not be part of the remote players repository at all.

The only way around this would be to query the pusher for the location of the user (I think we already do this for the "walk to" feature)

moufmouf (changes requested):

PR #5482, opened 2025-12-18
move personal desk actions in tools menu

ydiarra:


moufmouf (changes requested):
The "tools" menu visibility is managed by the `mapMenuVisibleStore`.
You need to change this store to make sure that if a user has a Desk, the menu will be displayed.

PR #5509, opened 2026-01-02
fix: Construct entity collection and file URLs relative to map-storage

BAWES:
Entity collection is stored at map-storage root, not relative to WAM file.

Previously, entity collection and file upload URLs were constructed relative to the WAM file path, causing errors when WAM and entities are on different paths (Admin API). Files are stored at the map-storage root, not relative to the WAM file location.

This fix:
- Constructs entity collection URL relative to map-storage origin
- Constructs file upload URLs relative to map-storage origin
- Works with both default setup and Admin API setup
- Works with USE_DOMAIN_NAME_IN_PATH=true or false
- Works with both S3 and local storage

### DIFF
--- a/play/src/front/Components/MapEditor/PropertyEditor/FileUpload/FileUpload.svelte
+++ b/play/src/front/Components/MapEditor/PropertyEditor/FileUpload/FileUpload.svelte
@@ -78,9 +78,14 @@
         const fileName = selectedFile.name.slice(0, lastDot);
         const fileExt = selectedFile.name.slice(lastDot + 1);
 
-        const fileUrl = `${get(gameSceneStore)?.room.mapStorageUrl?.toString()}private/files/${fileName}-${
-            property.id
-        }.${fileExt}`;
+        const mapStorageUrl = get(gameSceneStore)?.room.mapStorageUrl;
+        if (!mapStorageUrl) {
+            throw new Error("Map storage URL is not available");
+        }
+        // Files are stored at map-storage root, not relative to WAM file
+        // Construct URL relative to map-storage origin
+        const filePath = `private/files/${fileName}-${property.id}.${fileExt}`;
+        const fileUrl = new URL(filePath, `${mapStorageUrl.protocol}//${mapStorageUrl.host}`).toString();
 
         property.name = selectedFile.name;
         property.link = fileUrl;
--- a/play/src/front/Components/PopUp/PopUpDropFileEntity.svelte
+++ b/play/src/front/Components/PopUp/PopUpDropFileEntity.svelte
@@ -67,9 +67,15 @@
         const lastDot = file.name.lastIndexOf(".");
         const name = file.name.slice(0, lastDot);
         const fileExt = file.name.slice(lastDot + 1);
-        const fileUrl = `${get(
-            gameSceneStore
-        )?.room.mapStorageUrl?.toString()}private/files/${name}-${propertyId}.${fileExt}`;
+
+        const mapStorageUrl = get(gameSceneStore)?.room.mapStorageUrl;
+        if (!mapStorageUrl) {
+            throw new Error("Map storage URL is not available");
+        }
+        // Files are stored at map-storage root, not relative to WAM file
+        // Construct URL relative to map-storage origin
+        const filePath = `private/files/${name}-${propertyId}.${fileExt}`;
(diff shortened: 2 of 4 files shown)


moufmouf:
Hey @BAWES !

I wonder if the whole issue does not come from here:

https://github.com/workadventure/workadventure/blob/99bd97c9036a330c1a443e00ece355591314a508/play/src/front/Connection/Room.ts#L282-L282

room.mapStorageUrl is not supposed to point to the WAM file. It's supposed to point to the root of the map storage.

We build the path to the file like this:

```
const fileUrl = `${get(gameSceneStore)?.room.mapStorageUrl?.toString()}private/files/${fileName}-${
            property.id
        }.${fileExt}`;
```

In your case, my guess is that room.mapStorageUrl is incorrect.

It could be incorrect if PUBLIC_MAP_STORAGE_PREFIX (which is derived from PUBLIC_MAP_STORAGE_URL) is incorrect.

So before we take a deeper look at this PR, could you check the value of your PUBLIC_MAP_STORAGE_URL url (and maybe try to add or remove a slash at the end of the URL and see if this solves things)

BAWES:
Thanks @moufmouf for the feedback!
I made some adjustments to it based on your feedback but im not sure if its the correct approach but it seems to be resolving the issue we're facing.

The following explains the issue we were facing and how this resolves it.

## The Original Issue
**Problem:** Uploaded files and entities disappeared after page refresh with errors like "Could not find entity ... in collection custom entities".
**Root Cause:** URLs were constructed relative to the WAM file path instead of the map-storage root.
### Example:
WAM URL: http://host.docker.internal:3000/play.workadventure.localhost/test/test/test/map.wam
Old code: `new URL("assets/entities/entities.json", wamUrlFile)`
Result: http://host.docker.internal:3000/play.workadventure.localhost/test/test/test/assets/entities/entities.json ❌
Should be: http://host.docker.internal:3000/assets/entities/entities.json ✅

## The First Fix (commit f03010231a857a83e424969f5dcb5fd50aa331fa)
I fixed `getCustomEntityCollectionUrl()` to extract the origin from the WAM URL:
```
// BEFORE (WRONG):
get mapStorageUrl(): URL | undefined {
    const mapStoragePath = `${PUBLIC_MAP_STORAGE_PREFIX}`; // Could be "" (empty)
    return new URL(mapStoragePath, this._wamUrl); // Uses WAM URL as base
}
```
This worked for direct access, but there was still an issue with `Room.mapStorageUrl`.

## The Problem You Identified
`Room.mapStorageUrl` was incorrectly derived from the WAM URL:
```
// BEFORE (WRONG):
get mapStorageUrl(): URL | undefined {
    const mapStoragePath = `${PUBLIC_MAP_STORAGE_PREFIX}`; // Could be "" (empty)
    return new URL(mapStoragePath, this._wamUrl); // Uses WAM URL as base
}
```
**Issue:** When `PUBLIC_MAP_STORAGE_PREFIX` is empty (because `PUBLIC_MAP_STORAGE_URL` has no path), `new URL("", wamUrl)` resolves to the WAM file URL itself, not the map-storage root.

## The Hotfix (commit 6b5d62c21)
Fixed `Room.mapStorageUrl` to:
- Extract the origin (protocol + host) from the WAM URL
- Append `PUBLIC_MAP_STORAGE_PREFIX` when available (for reverse proxy setups)
- Return the map-storage root URL, not the WAM file URL
```
// AFTER (CORRECT):
get mapStorageUrl(): URL | undefined {
    const wamUrlObj = new URL(this._wamUrl);
    const origin = `${wamUrlObj.protocol}//${wamUrlObj.host}`;
    
    if (PUBLIC_MAP_STORAGE_PREFIX) {
        return new URL(PUBLIC_MAP_STORAGE_PREFIX, origin);
    }
    return new URL(origin);
}
```
Then updated `getCustomEntityCollectionUrl()` to use `room.mapStorageUrl`:
```
const mapStorageUrl = this.room.mapStorageUrl;
return new URL(entityCollectionPath, mapStorageUrl.toString());
```

## Result
- `room.mapStorageUrl` points to the map-storage root (not the WAM file)
- Works with direct access (no prefix)
- Works with reverse proxy setups (with path prefix like `/map-storage`)
- `getCustomEntityCollectionUrl()` uses the correct base URL

This ensures entity collection URLs are constructed correctly regardless of the setup.

PR #5515, opened 2026-01-05
Implement big brother feat

ydiarra:


back/src/Services/SocketManager.ts
```diff
@@ -215,6 +215,7 @@ export class SocketManager {
                     room.roomGroup ?? new URL(room.roomUrl).host,
                     room.roomUrl
                 ),
+                bigBrowserActivated: room.wamSettings?.megaphone?.bigBrowserActivated ?? false,
```
moufmouf: Maybe rename to "audienceVideoFeedbackActivated" ?

PR #5516, opened 2026-01-05
Guest Access Mode

gparant:


play/src/front/Phaser/Game/GameManager.ts
```diff
@@ -55,6 +58,16 @@ export class GameManager {
         this.chatVisibilitySubscription = initializeChatVisibilitySubscription();
     }
 
+    private isGuestMode(): boolean {
+        // Check if enableFastpass is enabled in metadata
+        const enableFastPass = (this.startRoom.metadata as { room?: { enableFastPass?: boolean } })?.room
+            ?.enableFastPass;
```
moufmouf: Instead of one "enableFastPass" entry, I think we should aim for more properties that match the entries in the EnvironmentVariables. And the opposite is also true, we should EnvironmentVariables matching the data from the AdminAPI.

Also, the "metadata" section is reserved for scripting API use only. Anything that is read by the WorkAdventure application should be strongly typed and part of the isMapDetailsData Zod object.

play/src/front/Connection/Room.ts
```diff
@@ -389,6 +395,21 @@ export class Room {
...
+        return this._defaultGuestTexture;
+    }
+
+    get guestNameAppendRandomNumbers(): boolean | undefined {
+        if (this._guestNameAppendRandomNumbers === undefined) {
+            return false;
+        }
+        return this._guestNameAppendRandomNumbers;
+    }
+
```
moufmouf, line 411: Here, we could also have a

- get useRandomGuestName(): boolean
- get useRandomGuestTexture(): boolean

This data should come from the pusher and either EnvironmentVariables (if using no admin API) or the AdminAPI (via the /map endpoint)

play/src/front/Phaser/Game/GameManager.ts
```diff
@@ -78,8 +91,54 @@ export class GameManager {
         const preferredAudioInputDeviceId = localUserStore.getPreferredAudioInputDevice();
         const preferredVideoInputDeviceId = localUserStore.getPreferredVideoInputDevice();
 
-        console.info("Preferred audio input device: " + preferredAudioInputDeviceId);
-        console.info("Preferred video input device: " + preferredVideoInputDeviceId);
+        // Guest mode
+        if (this.isGuestMode()) {
```
moufmouf, line 94: Can we skip the call to "isGuestMode()" here?

My idea is that we should be able to fine tune things. Some people might want a random name, but not a random woka, and ask for the webcam. Other users might want to ask for the name and woka, but skip the camera, etc...
So I would avoid putting everything in a big "if" block. We should aim for something more fine grained.

play/src/front/Connection/Room.ts
gparant: Provided of "GUEST_NAME_APPEND_RANDOM_NUMBERS" env variable from the pusher and was developed by the community.

play/src/front/Phaser/Game/GameManager.ts
gparant: Maybe you're right, if one day someone asks us, we could ask or apply changes with the community 💪

PR #5522, opened 2026-01-07
Fix view port when we follow user

gparant:


play/src/front/Phaser/Game/GameScene.ts
```diff
@@ -1352,6 +1352,10 @@ export class GameScene extends DirtyScene {
                 throw new Error('Cannot find player with ID "' + userId + '"');
             }
             player.updatePosition(moveEvent);
+            // If the camera is following the player, we need to update the viewport
+            if (this.cameraManager?.playerFollowing === player) {
+                this.sendViewportToServer();
+            }
```
moufmouf, line 1358: Do we still need this if we have an event that send the viewport to the server each time the camera moves?

play/src/front/Phaser/Game/MapEditor/Tools/ExplorerTool.ts
```diff
@@ -164,6 +169,8 @@ export class ExplorerTool implements MapEditorTool {
         this.scene.input.off("pointermove", this.pointerMoveHandler);
         this.scene.input.off("pointerup", this.pointerUpHandler);
         this.scene.input.off(Phaser.Input.Events.GAME_OUT, this.pointerUpHandler);
+        // Unsubscribe to camera updates
+        this.scene.getCameraManager().off(CameraManagerEvent.CameraUpdate, this.updateViewport.bind(this));
```
moufmouf: The "off" here will fail because `this.updateViewport.bind(this)` is not the same instance as the one line 275 when you do the "on".

You should instead store the value of `this.updateViewport.bind(this)` in a property of the class.

Alternatively, you can write the updateViewport this way:

```
private updateViewport = () => {
        this.scene.sendViewportToServer();
    }
```

This creates a closure that is always bound to "this".

It's JS nonsense but it's the way it is.

play/src/front/Phaser/Game/MapEditor/Tools/ExplorerTool.ts
```diff
@@ -126,6 +127,10 @@ export class ExplorerTool implements MapEditorTool {
         this.scene.markDirty();
     };
 
+    private updateViewport() {
+        this.scene.sendViewportToServer();
```
moufmouf: You should use `this.scene.throttledSendViewportToServer` instead of `this.scene.sendViewportToServer`. Otherwise, we are going to kill the server with one call for each frame. throttledSendViewportToServer does a throttling every 200ms.

play/src/front/Phaser/Game/GameScene.ts
gparant: In this case, we are following the remote player. To create a new viewport, the current player must take a action: move up, down...
The current player does not move during the following, the viewport will not be updated, resulting, we lost the remote player.

PR #5528, opened 2026-01-08
feat: Improve activation detection with rectangle-based distance calculation

gparant:
## 🎯 Objective

Improve the accuracy of entity activation detection by replacing point-based distance calculation with a rectangle collision system.

### Rectangle Distance Calculation

- **Added `MathUtils.distanceBetweenRectangles()`**: New function that calculates the minimum distance between two rectangles
  - Returns `0` if rectangles overlap or touch
  - Calculates the minimum Euclidean distance between rectangle edges otherwise

### Rectangle System for Entities

- **`Entity.getCollisionRectangle()`**: Returns the base collision rectangle of the entity
- **`Entity.getActivationRectangle()`**: Returns the activation rectangle (collision + `activationRadius` extended on all sides)

### Rectangle System for Characters

- **`Character.getCollisionRectangle()`**: Returns the character's collision rectangle based on Phaser physics body
  - Uses `body.left` and `body.top` to get the actual world position

### ActivatablesManager Updates

- **`updateActivatableObjectsDistances()`**: Now uses rectangles to calculate distances with entities
- **`updateDistanceForSingleActivatableObject()`**: Updated to use rectangles
- **`findNearestActivatableObject()`**: Now detects rectangle overlaps (distance = 0) instead of comparing with `activationRadius`

## 🐛 Fixes

- Fixed character center calculation to use the actual physics body position (`body.left`, `body.top`) instead of `this.x + body.offset.x`

Issue in video:

https://github.com/user-attachments/assets/3ef922a5-5bf2-45a8-9ea3-a9ce4048c2c8

Now with the fix:

https://github.com/user-attachments/assets/79dd97c3-1300-4bbb-b245-ebc99aded3b1

## 🧪 Testing

- Verified that entities are correctly detected when the player's rectangle overlaps the activation rectangle
- Validated that distance is correctly calculated when rectangles don't touch
- Tested debug visualization in development mode

### DIFF
--- a/play/src/front/Phaser/Game/ActivatablesManager.ts
+++ b/play/src/front/Phaser/Game/ActivatablesManager.ts
@@ -67,24 +67,14 @@ export class ActivatablesManager {
     }
 
     public deduceSelectedActivatableObjectByDistance(): void {
-        console.log("deduceSelectedActivatableObjectByDistance => this.canSelectByDistance", this.canSelectByDistance);
         if (!this.canSelectByDistance) {
             return;
         }
         const newNearestObject = this.findNearestActivatableObject();
-        console.log("deduceSelectedActivatableObjectByDistance => newNearestObject", newNearestObject);
-        console.log(
-            "deduceSelectedActivatableObjectByDistance => this.selectedActivatableObjectByDistance",
-            this.selectedActivatableObjectByDistance
-        );
         if (this.selectedActivatableObjectByDistance === newNearestObject) {
             return;
         }
         // update value but do not change the outline
-        console.log(
-            "deduceSelectedActivatableObjectByDistance => this.selectedActivatableObjectByPointer",
-            this.selectedActivatableObjectByPointer
-        );
         if (this.selectedActivatableObjectByPointer == newNearestObject) {
             this.selectedActivatableObjectByDistance = newNearestObject;
             return;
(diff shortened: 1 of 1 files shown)


moufmouf:
@gparant I think your changes broke the test "Map editor" > "interacting with object". This might need to be adjusted.

play/src/front/Phaser/Game/ActivatablesManager.ts
```diff
@@ -108,30 +109,46 @@ export class ActivatablesManager {
     }
 
     public updateActivatableObjectsDistances(objects: ActivatableInterface[]): void {
-        const currentPlayerPos = this.currentPlayer.getDirectionalActivationPosition(
-            this.directionalActivationPositionShift
-        );
+        const playerRect = this.currentPlayer.getCollisionRectangle();
```
moufmouf, line 112: I think we can slightly improve that. Here, we say the interaction is triggered if the collision rectangle of the player is close to the collision rectangle of the activable object.

But in most video games, the activation zone is not the player zone itself, but a zone right in front of the player. This way, when we look at an object 32px in front of us, it is activable, but when we turn the back to this object, it is no more activable (event if we are next to it).

So basically, we should move the "collision rectangle" of the player 16 to 32 px in the direction the user is facing.

Also, "collision rectangle" is not the correct name. We should call that "interaction rectangle" instead.

gparant: I'm not sure about the proposed user experience. The object is not a video game, but when the user moves around it.
If users are not using precise movements such as trackpad, right-click, joystick, mobile, or keyboard, then the object should remain active for them.

PR #5543, opened 2026-01-14
Fix some visual issues on the OICD profile screen

Abestanis:
The text was forced to white, but the background was not specified. So with a default white background from the browser, the text was only visible when selecting it (see screenshot). Also, the text `false` was inserted by mistakes for empty values.

The profile screen, with some selected white on white text


moufmouf:
Nice! Thanks a lot!

PR #5580, opened 2026-01-15
fix(follow): fix shortcut behavior and abort message propagation

ydiarra:
## Summary
- Fix the follow shortcut (F key) that was triggering twice, skipping the confirmation popup
- Fix the follow abort not being received by users who haven't accepted the request yet
- Ensure the follow popup closes immediately when receiving an abort message

## Problem
The follow feature had several issues:
1. **Double trigger**: The F shortcut was using `isDown` instead of `JustDown`, causing it to fire on every frame while held. This made the state jump from "off" → "active" → "ending" instantly, bypassing the waiting popup
2. **One-way abort**: When a leader canceled a follow request, only confirmed followers received the abort message. Users who received the request but hadn't accepted yet could still click "Accept"
3. **Stale popup**: The follow popup didn't close immediately when receiving an abort, creating a race condition

## Solution
- Use `Phaser.Input.Keyboard.JustDown()` for the Follow event to detect only the initial key press (edge trigger)
- Broadcast the abort message to the entire group when the leader cancels (same as the follow request)
- Explicitly remove the popup in `FollowManager` when receiving an abort message

### DIFF
--- a/play/src/front/Phaser/UserInput/UserInputManager.ts
+++ b/play/src/front/Phaser/UserInput/UserInputManager.ts
@@ -313,15 +313,8 @@ export class UserInputManager {
                     d.keyInstance.reset();
                     return;
                 }
-                // For Follow event, use JustDown to detect only the initial key press (edge trigger)
-                // This prevents the action from being triggered multiple times while the key is held
-                if (d.event === UserInputEvent.Follow) {
-                    if (Phaser.Input.Keyboard.JustDown(d.keyInstance)) {
-                        eventsMap.set(d.event, true);
-                    }
-                } else {
-                    eventsMap.set(d.event, true);
-                }
+
+                eventsMap.set(d.event, true);
             }
         });
         return eventsMap;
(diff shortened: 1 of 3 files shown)


play/src/front/Phaser/UserInput/UserInputManager.ts
```diff
@@ -313,7 +313,15 @@ export class UserInputManager {
...
-                eventsMap.set(d.event, true);
+                // For Follow event, use JustDown to detect only the initial key press (edge trigger)
+                // This prevents the action from being triggered multiple times while the key is held
+                if (d.event === UserInputEvent.Follow) {
+                    if (Phaser.Input.Keyboard.JustDown(d.keyInstance)) {
+                        eventsMap.set(d.event, true);
+                    }
+                } else {
+                    eventsMap.set(d.event, true);
+                }
```
moufmouf: This seems weird to put the code here. We should maybe try to put this where we put the other shortcuts? For instance the shortcut to open the chat or the map editor (that work in a similar way)

PR #5589, opened 2026-01-19
support nested speaker/listener zones

ydiarra:
Allow podium (speaker) zones to be placed inside attendees (listener) zones. When entering a nested zone, the system now switches roles instead of trying to rejoin the space.

- Add getCurrentSpace/getCurrentSpaceName methods to ProximityChatRoom
- Track active megaphone zones with MegaphoneZoneState
- Handle role transitions in speaker/listener enter/leave handlers
- Prevent duplicate joinSpace/leaveSpace calls for same space


moufmouf:
@codex review

PR #5591, opened 2026-01-20
Fix LockByKey race condition in concurrent requests

ydiarra:
## Summary
- Fix race condition where multiple concurrent requests for the same key would await the same promise and then execute in parallel when it resolves
- Implement promise chaining pattern: each new request chains after the previous one, ensuring strict sequential execution
- Add configurable timeout (default 10s) with Sentry error reporting

## Problem
When requests A, B, C arrive for the same key while A is running:
- **Before**: B and C both await `promiseA`, then execute in parallel when A finishes
- **After**: B chains after A, C chains after B → strict A → B → C execution

## Test plan
- [x] `should execute concurrent requests sequentially (not in parallel)`
- [x] `should chain requests so B and C don't run in parallel when both arrive during A`
- [x] `should allow parallel execution for different keys`
- [x] `should auto reject after timeout`
- [x] `should free the lock when promise is rejected`

### DIFF
--- a/map-storage/src/Services/LockByKey.ts
+++ b/map-storage/src/Services/LockByKey.ts
@@ -4,6 +4,7 @@
  * ensuring sequential execution even when multiple requests arrive concurrently.
  */
 import * as Sentry from "@sentry/node";
+import { asError } from "catch-unknown";
 
 export class LockByKey<T> {
     private locks = new Map<T, Promise<void>>();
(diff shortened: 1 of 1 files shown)


map-storage/src/Services/LockByKey.ts
```diff
@@ -1,31 +1,73 @@
...
+            }, timeoutMs);
+
+            promise
+                .then((result) => {
+                    clearTimeout(timeoutHandle);
+                    resolve(result);
+                })
+                .catch((err: unknown) => {
+                    clearTimeout(timeoutHandle);
+                    reject(err instanceof Error ? err : new Error(String(err)));
```
moufmouf:
```suggestion
                    reject(asError(err));
```

PR #5594, opened 2026-01-20
Add FPS standard deviation to WebRTC stats

moufmouf:
## Summary

- track FPS sample standard deviation in WebRTC stats (compute standard deviation on the last 8 FPS sample)
- reset FPS window on resolution changes (because FPS could change when resolution change)
- color stats box based on FPS std dev


moufmouf:
@Codex please review

PR #5599, opened 2026-01-21
Update GlobalCommunicationModal to use reactive stores for camera and microphone device IDs

ydiarra:
- Replaced local variables for camera and microphone device IDs with reactive store subscriptions.
- Improved state management for device selection in the modal component.

### DIFF
--- a/play/src/front/Components/Modal/GlobalCommunicationModal.svelte
+++ b/play/src/front/Components/Modal/GlobalCommunicationModal.svelte
@@ -15,6 +15,8 @@
         requestedMicrophoneState,
         streamingMegaphoneStore,
         localStreamStore,
+        usedCameraDeviceIdStore,
+        usedMicrophoneDeviceIdStore,
     } from "../../Stores/MediaStore";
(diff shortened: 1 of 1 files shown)


play/src/front/Components/Modal/GlobalCommunicationModal.svelte
```diff
@@ -466,7 +468,7 @@
                                     alt="Turn off microphone"
                                 />
                                 <div class="w-full">
-                                    <Select bind:value={cameraDeviceId} onChange={() => selectCamera()}>
+                                    <Select bind:value={$usedCameraDeviceIdStore} onChange={() => selectCamera()}>
```
moufmouf: In theory, we should not set the "usedCameraDeviceIdStore". The "usedCameraDeviceIdStore" should only be set by the MediaStore when we successfully set the value after managing to start the the camera/microphone.

We should really use a temporary cameraDeviceId here that is both updated when we select the camera, but also when the usedCameraDeviceIdStore changes.

moufmouf (changes requested):

PR #5617, opened 2026-01-26
Use MINIMUM_DISTANCE on the frontend for conversation bubble rendering

### LINKED ISSUE #5537: Allow changing the bubble size
Hello, thank you for this amazing software. I have it deployed on my own server and I would like to be able to change the default bubble size for the proximity voice chat.

At first glance it looks like I would just need to change the following line:
https://github.com/workadventure/workadventure/blob/25a8ad14692399ffe51a55bdec51268c8d8d48dd/play/src/front/Phaser/Entity/ConversationBubble.ts#L18

Would you accept a PR that allows to overwrite that constant with an environment variable on the server?

Or is there some other way to change the bubble size that I missed?

Abestanis:
It was already possible to change the conversation bubble size on the backend using the `MINIMUM_DISTANCE` and `GROUP_RADIUS` environment variables, although they were hard to discover. But the frontend did not take these variables into account, so the rendered bubble would stay the same even if the server side bubble area increased. This makes it so the frontend bubble rendering takes the `MINIMUM_DISTANCE` variable into account. Fixes #5537.

| Before | After |
|--------|--------|
| Bubble is small in-between two avatars | Bubble is large |


moufmouf:
Thanks a lot @Abestanis :+1:

PR #5650, opened 2026-02-03
Display share fields through a tooltip element

HugoAverty:
instead of settings tabs, removing settings share tab


moufmouf:
You can probably improve a bit the design regarding spacing here

image

PR #5651, opened 2026-02-03
feat(map-storage): Allow modifyAreaMessage without canEdit for personal area claim/revoke only

ydiarra:
## Problem

Users without map edit rights (`userCanEdit`) could not claim or revoke personal areas, because `modifyAreaMessage` was rejected by the generic permission check. Personal area claim/revoke only updates `personalAreaPropertyData.ownerId` (and optionally the area name) and should be allowed for any user on unclaimed areas or on their own claimed area.

## Solution

- Add `modifyAreaMessage` to `COMMANDS_ACCESSIBLE_WITHOUT_CAN_EDIT` so the generic check no longer blocks it.
- Introduce `isModifyAreaMessageOnlyClaim(message, userUUID, existingArea)` to ensure that when the user does not have `userCanEdit`, the command is **only** a valid claim or revoke:
  - **Claim**: existing area has `ownerId === null`, command sets `ownerId` to `userUUID`, geometry and other properties unchanged.
  - **Revoke**: existing area has `ownerId === userUUID`, command sets `ownerId` to `null`, geometry and other properties unchanged.
- In the `modifyAreaMessage` case, when `!userCanEdit`, call this verifier before applying the command; if it returns `false`, throw so the client receives an error command.

## Changes

- **map-storage/src/MapStorageServer.ts**
  - Add `"modifyAreaMessage"` to `COMMANDS_ACCESSIBLE_WITHOUT_CAN_EDIT`.
  - Add `ModifyAreaMessagePayload` type and `isModifyAreaMessageOnlyClaim()` (checks existing area state + command contains only claim-related data).
  - In `case "modifyAreaMessage"`: when `!userCanEdit`, validate with `isModifyAreaMessageOnlyClaim()` and reject if not a valid claim/revoke.
  - Replace previous TODOs with a short comment and a TODO for future improvement (e.g. dedicated personal area data or WAM section).

### DIFF
--- a/map-storage/src/MapStorageServer.ts
+++ b/map-storage/src/MapStorageServer.ts
@@ -22,7 +22,6 @@ import type {
     EditMapCommandsArrayMessage,
     EditMapCommandWithKeyMessage,
     MapStorageClearAfterUploadMessage,
-    ModifyAreaMessage,
     PingMessage,
     UpdateMapToNewestWithKeyMessage,
 } from "@workadventure/messages";
@@ -42,6 +41,7 @@ import { DeleteEntityMapStorageCommand } from "./Commands/Entity/DeleteEntityMap
 import { UploadFileMapStorageCommand } from "./Commands/File/UploadFileMapStorageCommand";
 import { hookManager } from "./Modules/HookManager";
 import { UpdateEntityMapStorageCommand } from "./Commands/Entity/UpdateEntityMapStorageCommand";
+import { isModifyAreaMessageOnlyClaim } from "./Services/isModifyAreaMessageOnlyClaim";
 
 const editionLocks = new LockByKey<string>();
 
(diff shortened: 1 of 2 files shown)


map-storage/src/MapStorageServer.ts
```diff
@@ -135,6 +213,8 @@ const mapStorageServer: MapStorageServer = {
 
                 const commandId = editMapCommandMessage.id;
 
+                // TODO: Remove this check once we have a proper system for associating data with areas
+                // (e.g., personal area data) or a WAM section accessible to all users.
```
moufmouf:
```suggestion
```

map-storage/src/MapStorageServer.ts
```diff
@@ -56,8 +57,85 @@ const COMMANDS_ACCESSIBLE_WITHOUT_CAN_EDIT = new Set<string>([
...
+/**
+ * Returns true iff:
+ * 1. The existing area has a personal area property and is in a valid state for claim/revoke
+ *    (for claim: not already claimed, i.e. ownerId === null; for revoke: claimed by this user, i.e. ownerId === userUUID).
+ * 2. The command contains only claim-related data (geometry unchanged, only personalAreaPropertyData.ownerId
+ *    set to userUUID or null, and optionally name).
+ */
+// TODO: Remove this check once we have a proper system for associating data with areas
+// (e.g., personal area data) or a WAM section accessible to all users.
+function isModifyAreaMessageOnlyClaim(
```
moufmouf: Let's move this to another file in order to keep the main file readable.

moufmouf: Also, let's put a big comment explaining what this does and why it is needed

PR #5671, opened 2026-02-06
WIP : Livekit Speech-to-text

ydiarra:


moufmouf:
We will need to modify the architecture of this app.

According to https://docs.livekit.io/deploy/custom/deployments/ , LiveKit recommends 4 cores and 8GB per agent server as a starting rule for most voice AI apps. This agent server can handle 10-25 concurrent jobs, depending on the components in use.

If we put the agent in the back component, suddenly, the back container will need much more CPU and anyway, will be limited to a few conversations.

We need to externalize the agent in a separate container, that we can put in an autoscaling HPA.

moufmouf:
An AgentSession only transcribes one given participant (see https://docs.livekit.io/agents/logic/sessions/#linked-participant )
We will need to attach one AgentSession per participant.

PR #5686, opened 2026-02-13
refactor: Integrate ListenerBox into CamerasContainer with shared ResizeHandle

ydiarra:
## Problem

ListenerBox was rendered separately in PresentationLayout with its own resize logic. This led to:
- Duplicated resize behavior
- Different layout/CSS from VideoBoxes
- Container collapsing to a thin bar when `max-height: 25%` was applied in one-line mode
- ListenerBox becoming too small when the user moved around the map

## Solution

- Move ListenerBox into CamerasContainer and reuse the existing ResizeHandle
- Remove ListenerBox’s custom resize logic and make it a simple content component
- Style ListenerBox like VideoBoxes (same `camera-box` wrapper and layout)
- Adjust PresentationLayout so the container gets proper height when showing ListenerBox
- Add minimum sizes: 320px when showing video, 360×220px when showing “waiting for speaker”

## Changes

- **CamerasContainer.svelte**: Add `showListenerBox` logic, render ListenerBox with VideoBox-style wrapper, use ResizeHandle for both modes, compute dimensions with minimums for normal and waiting states
- **PresentationLayout.svelte**: Remove direct ListenerBox usage, extend cameras container condition to include listener mode, disable `max-height-quarter` and enable `h-full`/`flex-1` when `showListenerBox`
- **ListenerBox.svelte**: Remove resize logic and props, accept `width`/`height` from parent, align structure with MediaBox
- **MediaStore.ts**: Add `removeTrack()` when stopping video/audio tracks (cleanup)


moufmouf:
Stupid question but I'll ask: could we not simplify that even further by making the ListenerBox a specific case of Streamable? (like the videos started from the scripting API)?

PR #5697, opened 2026-02-17
fix: Correct screen-block handling and recenter woka after follow/locate

ydiarra:
## Problem

- When stopping to follow a remote player (e.g. after closing the woka menu), the camera did not reposition, so the local woka was not recentered with respect to the current screen-block (discussion box / available area).
- After the "follow remote player" tween ended, the camera did not switch back to follow mode and did not run repositioning, so the woka could stay off-center relative to the screen-block.
- The woka menu store subscription reacted to the initial value on subscribe, causing unwanted follow/stop-follow actions when the store was first read.

## Solution

- Restore follow mode and run repositioning when the follow-remote-player tween completes, and when stopping to follow a remote player, so the camera offset is recomputed from the current biggest available area (screen-block).
- Add a dedicated `setFollowMode()` that sets the camera to follow mode and calls `reposition()` for a single place to "follow + respect screen-block".
- When a locate response is applied, call `setFollowMode()` after moving the camera so the woka is recentered relative to the screen-block.
- In `LocateManager`, ignore the first emission of the woka menu store so we only react to real open/close changes, not the initial value.

## Changes

- **CameraManager.ts**: Call `setFollowMode()` at the end of the follow-remote-player tween; call `this.scene.reposition()` in `stopFollowRemotePlayer()`; add `setFollowMode()` and use it where follow mode + reposition must be restored. (Note: remove any temporary `console.trace` / `console.log` before merge.)
- **GameScene.ts**: Trigger camera offset update when the biggest available area is recomputed (existing behavior); remove temporary debug logs before merge if present.
- **LocateManager.ts**: Skip first emission in woka menu store subscription; call `cameraManager.setFollowMode()` after moving the camera on locate so the woka is recentered with respect to the screen-block.

### DIFF
--- a/play/src/front/Phaser/Game/LocateManager.ts
+++ b/play/src/front/Phaser/Game/LocateManager.ts
@@ -40,7 +40,6 @@ export class LocateManager {
 
     private subscribeToWokaMenuStore(): void {
         // Subscribe to woka menu store to stop following the remote player when the woka menu is closed.
-        // Skip the first emission (current value at subscribe time) so we only react to actual changes.
         this.wokaMenuStoreUnsubscriber = wokaMenuStore.subscribe((value) => {
             if (value === undefined) {
                 this.cameraManager.stopFollowRemotePlayer();


play/src/front/Phaser/Game/LocateManager.ts
```diff
@@ -39,10 +39,10 @@ export class LocateManager {
     }
 
     private subscribeToWokaMenuStore(): void {
-        // Subscribe to woka menu store to stop following the remote player when the woka menu is closed
+        // Subscribe to woka menu store to stop following the remote player when the woka menu is closed.
+        // Skip the first emission (current value at subscribe time) so we only react to actual changes.
```
moufmouf:
```suggestion
```

PR #5698, opened 2026-02-17
refactor: improve back typescript configuration

nolway:
Improve the backend service’s TypeScript configuration by adding stronger typing and enforcing proper error throwing instead of having silent failures.


back/src/Model/Space.ts
```diff
@@ -317,10 +317,9 @@ export class Space implements CustomJsonReplacerInterface, ICommunicationSpace {
                 return user.megaphoneState || user.attendeesState;
             }
             default: {
-                const _exhaustiveCheck: never = this._filterType;
+                throw new Error(`Unknown filter type`);
```
moufmouf, line 320: Ouch! That's one hell of a PR. Thanks Alexis!

I'm not sure about those changes.
Here, you are trading a compile time error for a runtime error.
The compile time error seems better to me (even though that means leaving an unused variable).
The best thing would of course be to have a compile time error without the unused var but I'm not sure how to do this.

Maybe we could try something like:

```
function thisShouldNeverBeCalled(variable: never) {}
```

and

```
default: {
    thisShouldNeverBeCalled(this._filterType)
}
```

?

moufmouf (changes requested):

PR #5755, opened 2026-02-24
Invite meeting

gparant:
# Invite User
Button to invite user to meet us or join meeting.

# Create participants list
We expose the list of people in the current meeting via currentMeetingParticipantsStore on ProximityChatRoom (in play/src/front/Chat/Connection/Proximity/ProximityChatRoom.ts).

In WorkAdventure, “being in a meeting” is implemented as being in a space:
A space is the logical container for a given meeting (one bubble, one meeting room on the map, etc.).
ProximityChatRoom is the chat room tied to that space and represents the current conversation/meeting for the user.
So: The “current meeting” = the current space.
“Meeting participants” = users in that space.
Those users are already tracked by ProximityChatRoom via _space and _space.usersStore.
By putting currentMeetingParticipantsStore on ProximityChatRoom, we keep a single place that both “is the current meeting” and “knows who is in it.” That avoids duplicating logic and keeps the participant list aligned with the same concept of “meeting” used for chat, sounds, etc.

We use currentMeetingParticipantsStore on ProximityChatRoom because:
ProximityChatRoom already represents the current meeting (space) and has access to its user list.
Exposing a dedicated store gives a single, reactive, UI-ready list of participants.
It keeps one clear definition of “who is in the meeting” and avoids duplicated or divergent logic.
So the choice is both an architectural one (one owner for “current meeting” and “its participants”) and a practical one (simple, reactive API for any UI that needs the participant list).


docs/user/chat.md
```diff
@@ -98,6 +98,49 @@ them with a "green" or "orange" dot in the **Disconnected users** section.
...
+When you send an invitation, a sound notification will play to confirm the invitation was sent.
+
+### Receiving meeting invitations
+When another user invites you to their meeting, you will see a popup notification with:
+
+The name of the user inviting you
+An "Accept" button to join the meeting
+A "Decline" button to reject the invitation
+If you accept the invitation, you will automatically move to join the meeting with the inviting user.
+
```
moufmouf, line 120: Screenshots would be great here.

docs/user/chat.md
```diff
@@ -98,6 +98,49 @@ them with a "green" or "orange" dot in the **Disconnected users** section.
...
+Your own avatar
+Avatars of other participants in the meeting
+A count of total participants (if there are more than 3 participants)
+
+From the participants menu, you can:
+- Click on a participant to open their profile menu
+- Send a message to the proximity chat
+- Invite additional users to the meeting
+
```
moufmouf: This is unrelated to the chat. The current doc is for the chat. Maybe move this to a "meetings.md" doc?

play/src/front/Components/MeetingInvitation/MeetingInvitationAcceptToast.svelte
moufmouf, line 1: This toast is displayed to the wrong user (it is shown to the user that accepts the invite and not the user that sent the invite).

docs/user/chat.md
```diff
@@ -98,6 +98,49 @@ them with a "green" or "orange" dot in the **Disconnected users** section.
...
+### Receiving meeting invitations
+When another user invites you to their meeting, you will see a popup notification with:
+
+The name of the user inviting you
+An "Accept" button to join the meeting
+A "Decline" button to reject the invitation
+If you accept the invitation, you will automatically move to join the meeting with the inviting user.
+
+### Viewing meeting participants
+You can view the list of participants currently in your meeting by clicking on the participants icon in the action bar. This menu shows:
```
moufmouf: Probably belongs to another PR, right?

play/src/front/Components/MeetingInvitation/MeetingInvitationAcceptToast.svelte
gparant: In cause we use the user uuid because in the chat we don't have the user id.

gparant:
> You should use player id instead of uuid (we target a woka, not a user)

We use the user uuid because in the chat we don't have the user id.
In this case, we need to notify all users with the same uuid connected.

PR #5758, opened 2026-02-24
fix: hide listener screen share in megaphone see-attendees space

ydiarra:
## Problem

In a megaphone space with "see attendees" (LIVE_STREAMING_USERS_WITH_FEEDBACK), the speaker could see listeners' screen shares. Listeners should be able to activate screen share (button, capture), but in that mode only the speaker's screen share should be published and visible; listeners must not send or display their screen share to others.

## Solution

- **Publish/send control**: In see-attendees spaces, screen share is only published (LiveKit) or sent (WebRTC) when the local user is a speaker (`megaphoneState === true`). The condition is centralized in `Space.shouldPublishScreenShare()` and used by LiveKitRoom and SimplePeer; `syncScreenSharePublishState(shouldPublish?)` is triggered from Space (updateUserData, startStreaming, stopStreaming) via SpacePeerManager so the peer reacts immediately even before the server round-trip.
- **UI/boxes**: Screen share boxes are only created for users we are allowed to see: in see-attendees, only speakers’ screen share is shown (`shouldShowRemoteUserScreenShare` in Space). When a remote user’s `megaphoneState` changes, we create or remove their screen share box accordingly in `updateUserData`.

## Changes

- **Space / SpaceInterface**: `shouldPublishScreenShare()` and `shouldShowRemoteUserScreenShare(user)`; call `_peerManager.syncScreenSharePublishState(true|false)` from startStreaming/stopStreaming and `syncScreenSharePublishState()` from updateUserData when local user’s megaphoneState changes; create screen share box only when `shouldShowRemoteUserScreenShare` (initUsers, addUser, updateUserData); on remote megaphoneState change, remove listener’s screen share box or create it when they become speaker and share screen.
- **SpacePeerManager**: `syncScreenSharePublishState(shouldPublish?: boolean)` on manager and ICommunicationState/SimplePeerConnectionInterface, implemented in LivekitState, WebRTCState, DefaultCommunicationState.
- **LivekitConnection, LiveKitRoomInterface, LiveKitRoom**: `syncScreenSharePublishState(shouldPublish?)`; use param when provided, else `space.shouldPublishScreenShare()`.
- **SimplePeer**: `syncScreenSharePublishState(shouldPublish?)` with same logic; guards with `shouldPublishScreenShare()` when subscribing to stream, on peer connect, and in receiveWebrtcScreenSharingSignal.
- **Result**: Listeners can activate screen share but do not publish/send it; only the speaker’s screen share is visible; switching speaker/listener updates publish state and screen share boxes reactively.

### DIFF
--- a/play/src/front/Livekit/LiveKitRoom.ts
+++ b/play/src/front/Livekit/LiveKitRoom.ts
@@ -56,6 +56,7 @@ export class LiveKitRoom implements LiveKitRoomInterface {
     private localScreenSharingAudioTrack: LocalAudioTrack | undefined;
     private localCameraTrack: LocalVideoTrack | undefined;
     private localMicrophoneTrack: LocalAudioTrack | undefined;
+    private readonly effectiveScreenSharingLocalStreamStore: Readable<LocalStreamStoreValue | undefined>;
     private screenShareUpdateQueue: Promise<void> = Promise.resolve();
     private unsubscribers: Unsubscriber[] = [];
     private rxjsSubscriptions: Subscription[] = [];
@@ -81,6 +82,10 @@ export class LiveKitRoom implements LiveKitRoomInterface {
         },
         private _localStreamStore: Readable<LocalStreamStoreValue> = localStreamStoreForPublishing
     ) {
+        this.effectiveScreenSharingLocalStreamStore = deriveSwitchStore(
+            this.screenSharingLocalStreamStore,
+            this.space.shouldPublishScreenShareStore
+        );
         this._livekitRoomCounter.increment();
     }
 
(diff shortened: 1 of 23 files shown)


play/src/front/Livekit/LiveKitRoom.ts
```diff
@@ -380,84 +385,179 @@ export class LiveKitRoom implements LiveKitRoomInterface {
...
+    }
 
-                await this.localParticipant.publishTrack(this.localScreenSharingAudioTrack, {
-                    source: Track.Source.ScreenShareAudio,
+    /**
+     * Syncs screen share publish state with megaphone role: publish if speaker in see-attendees space,
+     * unpublish if listener. Called when stream or megaphoneState changes.
+     * When shouldPublish is provided (e.g. from startStreaming/stopStreaming), use it; otherwise use space.shouldPublishScreenShare().
+     */
+    async syncScreenSharePublishState(shouldPublish?: boolean): Promise<void> {
```
moufmouf: We can probably simplify this:

- syncScreenSharePublishState should only set a local boolean store
- we could derive the localScreenshareStore from the one we get from WorkAdventure + syncScreenSharePublishState  (using a deriveSwitchStore )

PR #5772, opened 2026-02-25
fix: improve ui ux for recording list

gparant:
Screenshot 2026-02-25 at 19 27 43


gparant:
Fix this ui dropdown

Screenshot 2026-02-25 at 19 33 57

may could replace the content when we click on the dropdown

PR #5775, opened 2026-02-26
fix: clear screen share track refs on unpublish so restart uses publishTrack in LiveKit

ydiarra:
## Problem

When a user was in LiveKit (e.g. after a 4th user joined), stopped screen sharing, then restarted it, other participants saw "ScreenShare en cours de connexion" indefinitely. The stream never appeared because after `unpublishTrack()`, the code kept references to the old `LocalVideoTrack` / `LocalAudioTrack`. On restart, the logic took the `replaceTrack()` path instead of `publishTrack()`. `replaceTrack()` only updates the media source on an **already published** track; it does not re-publish. Since the track had been unpublished, no new publication was created and viewers never received a stream.

## Solution

In `unpublishAllScreenShareTrack()`, set `localScreenSharingVideoTrack` and `localScreenSharingAudioTrack` to `undefined` after unpublishing. The next time the user starts screen share, the code then takes the `publishTrack()` path, creating a new publication so viewers receive the stream correctly.

## Changes

- **play/src/front/Livekit/LiveKitRoom.ts**: Update comment in `unpublishAllScreenShareTrack()` to document why refs are cleared (so restart uses `publishTrack()`; the two `= undefined` lines were already present).
- **tests/tests/screensharing.spec.ts**: Add E2E test "LiveKit: stop then restart screen share" (3 users → 4th user triggers LiveKit, stop/restart share, assert Bob/Eve/Mallory see Alice and that a real video stream is present via `#highlighted-media video` with `readyState >= 2` and `videoWidth > 0`).

### DIFF
--- a/play/src/front/Livekit/LiveKitRoom.ts
+++ b/play/src/front/Livekit/LiveKitRoom.ts
@@ -355,8 +355,10 @@ export class LiveKitRoom implements LiveKitRoomInterface {
                         Sentry.captureException(err);
                     });
                 }
-
-                if (streamResult) {
+                
+                if (!streamResult) {
+                    return;
+                }
                     // Create a new video track instance
                     const screenShareVideoTrack = streamResult.getVideoTracks()[0];
                     const screenShareAudioTrack = streamResult.getAudioTracks()[0];
@@ -433,8 +435,7 @@ export class LiveKitRoom implements LiveKitRoomInterface {
                                 Sentry.captureException(err);
                             });
                         }
-                    }
-                }
+                    } 
             })
         );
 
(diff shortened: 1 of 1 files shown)


play/src/front/Livekit/LiveKitRoom.ts
```diff
@@ -483,16 +483,17 @@ export class LiveKitRoom implements LiveKitRoomInterface {
...
                 })
         );
 
-        // Note: we don't clear local track references because of the memory leak issue mentioned above.
-        // We need to keep them to be able to replace the tracks when publishing a new screen share.
-        //this.localScreenSharingVideoTrack = undefined;
-        //this.localScreenSharingAudioTrack = undefined;
+        // Clear local track references so the next screen share uses publishTrack() instead of replaceTrack().
+        // After unpublish, the track is no longer published in the room; replaceTrack() does not re-publish.
+        // Without clearing refs, viewers would stay on "ScreenShare en cours de connexion" when the user restarts sharing.
```
moufmouf: Put the message in English here

play/src/front/Livekit/LiveKitRoom.ts
```diff
@@ -483,16 +483,17 @@ export class LiveKitRoom implements LiveKitRoomInterface {
                     if (track) {
                         // Note: for some reason, unpublishing / publishing a new track causes memory leaks.
                         // Instead, we just pause the upstream of the track when unpublishing, and "replaceTrack" when publishing a new one.
```
moufmouf: Comment to modify

PR #5826, opened 2026-03-09
fix(chat): clear chat search bar when closing chat

ydiarra:
## Problem

When the user closed the chat, the search bar kept its previous value. Reopening the chat showed the old search term, which was confusing and could affect filtering/behavior.

## Solution

Clear the chat search bar whenever the chat is hidden. The existing `initializeChatVisibilitySubscription` in `ChatStore` already reacts to chat visibility; we now reset `chatSearchBarValue` to an empty string when the chat becomes not visible, so the search bar is blank next time the chat is opened.

## Changes

- **`play/src/front/Chat/Stores/ChatStore.ts`**: In `initializeChatVisibilitySubscription`, when `chatVisibilityStore` becomes not visible, call `chatSearchBarValue.set("")` so the search bar is cleared on close.

### DIFF
--- a/play/src/front/Chat/Components/ChatHeader.svelte
+++ b/play/src/front/Chat/Components/ChatHeader.svelte
@@ -1,6 +1,6 @@
 <script lang="ts">
+    import { onDestroy } from "svelte";
     import { chatInputFocusStore } from "../../Stores/ChatStore";
-
     let searchActive = false;
     import { chatSearchBarValue, navChat, joignableRoom } from "../Stores/ChatStore";
     import LoadingSmall from "../images/loading-small.svelte";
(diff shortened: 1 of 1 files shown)


play/src/front/Chat/Stores/ChatStore.ts
```diff
@@ -54,6 +54,10 @@ export const chatSearchBarValue = writable<string>("");
 
 export function initializeChatVisibilitySubscription() {
     const unsubscriber = chatVisibilityStore.subscribe((visible) => {
+        if (!visible) {
+            chatSearchBarValue.set("");
+        }
+
```
moufmouf: That's a store mutating another store. Maybe not the easiest to track.

Wouldn't it make more sense to call `chatSearchBarValue.set("");` in the onDestroy of ChatHeader.svelte?

PR #5835, opened 2026-03-09
fix(chat): set Matrix room alias only for public rooms and make it unique

ydiarra:
## Problem

- Room alias (`room_alias_name`) was sent for every room and folder, so two rooms with the same name caused an alias conflict (Matrix requires alias uniqueness per server).
- This effectively prevented creating multiple rooms with the same name, although the Matrix spec allows duplicate room names.

## Solution

- Send `room_alias_name` only when `visibility === "public"` (for both rooms and folders), matching the spec that aliases are mainly for public rooms.
- For public rooms and folders, generate a unique alias with `uniqueRoomAliasName()` (slug + UUID) so multiple public rooms with the same name can coexist without alias conflicts.

## Changes

- **`play/src/front/Chat/Connection/Matrix/MatrixChatConnection.ts`**
  - `mapCreateRoomOptionsToMatrixCreateRoomOptions`: add `room_alias_name` only if `roomOptions.visibility === "public"`, using `uniqueRoomAliasName(roomName)`.
  - `mapCreateRoomOptionsToMatrixCreateFolderOptions`: same for folders (spaces).
  - New helper `uniqueRoomAliasName(baseName)`: returns `slugify(baseName) || "room"` + `-` + `crypto.randomUUID().replace(/-/g, "")` for a unique alias local part.
  - Comment added for future MSC4362 (encrypted state events) and optional `setRoomName` after creation.

### DIFF
--- a/play/src/front/Chat/Connection/Matrix/MatrixChatConnection.ts
+++ b/play/src/front/Chat/Connection/Matrix/MatrixChatConnection.ts
@@ -31,16 +31,8 @@ import { MapStore } from "@workadventure/store-utils";
 import { KnownMembership } from "matrix-js-sdk/lib/@types/membership";
 import { defaultWoka } from "@workadventure/shared-utils";
 import { slugify } from "@workadventure/shared-utils/src/Jitsi/slugify";
+import { shortHash } from "@workadventure/shared-utils/src/String/shortHash";
 
-/**
- * Returns a unique room alias local part (no domain) for public rooms/folders.
- * Ensures multiple rooms with the same name get distinct aliases.
- */
-function uniqueRoomAliasName(baseName: string): string {
-    const slug = slugify(baseName) || "room";
-    const uuid = crypto.randomUUID().replace(/-/g, "");
-    return `${slug}-${uuid}`;
-}
 import { AvailabilityStatus } from "@workadventure/messages";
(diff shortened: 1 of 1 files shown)


play/src/front/Chat/Connection/Matrix/MatrixChatConnection.ts
```diff
@@ -30,6 +30,16 @@ import * as Sentry from "@sentry/svelte";
...
 import { KnownMembership } from "matrix-js-sdk/lib/@types/membership";
 import { slugify } from "@workadventure/shared-utils/src/Jitsi/slugify";
+
+/**
+ * Returns a unique room alias local part (no domain) for public rooms/folders.
+ * Ensures multiple rooms with the same name get distinct aliases.
+ */
+function uniqueRoomAliasName(baseName: string): string {
+    const slug = slugify(baseName) || "room";
+    const uuid = crypto.randomUUID().replace(/-/g, "");
```
moufmouf: An alias is supposed to be easy to type. If we replace the ID of the room (a very long number) with an alias that is some string plus a very long number, we lose the interest of the alias.

Maybe we could try generating shorter random characters? like 6 random characters at the end of the name (we do such a thing for Jitsi room names)

moufmouf (changes requested):

PR #5839, opened 2026-03-10
Replace lodash with deepmerge-ts

angerenage:
Replace cjs lodash dependency with esm deepmerge-ts.

### DIFF
--- a/play/src/front/Phaser/Components/MapEditor/AreaPreview.ts
+++ b/play/src/front/Phaser/Components/MapEditor/AreaPreview.ts
@@ -1,4 +1,4 @@
-import { deepmergeInto } from "deepmerge-ts";
+import { deepmergeIntoCustom, type DeepMergeLeafURI } from "deepmerge-ts";
 import type {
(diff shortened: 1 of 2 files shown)


play/src/front/Phaser/Components/MapEditor/AreaPreview.ts
```diff
@@ -171,12 +171,18 @@ export class AreaPreview extends Phaser.GameObjects.Rectangle {
...
-                return;
+            const mergeProperty = deepmergeIntoCustom({
+                mergeArrays: (target, arrays) => {
+                    target.value.length = 0;
+
+                    const last = arrays[arrays.length - 1];
+                    if (last !== undefined) {
+                        target.value.push(...last);
+                    }
+                },
```
moufmouf: Seems to be different from the _.mergeWith.
It seems _.mergeWith  does not go deep into the arrays and simply takes the array of the target and applies it. deepmergeIntoCustom should do the same I guess.

PR #5841, opened 2026-03-11
Added setStatus event from map scripting

beingeek:
#### Intro:

PR introduces a new scripting API feature allowing the player's availability status to be set via `WA.player.setStatus()`. It adds support for statuses such as "ONLINE", "BUSY", "DO_NOT_DISTURB", and "BACK_IN_A_MOMENT", while ensuring that "SILENT" and "AWAY" remain auto-managed and cannot be set directly. The implementation includes type validation, event handling, and comprehensive tests to verify correct behavior.

#### New player status scripting API:
* Added `isSetStatusEvent` schema and `SetStatusEvent` type for validating and handling status change events in `SetStatusEvent.ts`.
* Extended the `IframeEvent` union and event imports to support the new "setStatus" event in `IframeEvent.ts`. 
* Implemented the `setStatus` method in `WorkadventurePlayerCommands` to send status change requests from scripts, including documentation and restrictions for auto-managed statuses.

#### Status event handling and integration
* Updated `IframeListener` to handle "setStatus" events, mapping allowed statuses to internal types and warning on attempts to set auto-managed statuses.
* Added necessary imports for status management logic in `IframeListener.ts`.

#### Testing Case:
* Introduced end-to-end tests in `scripting_setStatus.spec.ts` to verify correct status changes and rejection of auto-managed statuses via the scripting API.

### DIFF
--- a/docs/developer/map-scripting/references/api-player.md
+++ b/docs/developer/map-scripting/references/api-player.md
@@ -492,6 +492,31 @@ browsers automatically).
 
 ![](../images/outlines.png)
 
+## Set the status of the player
+```
+WA.player.setStatus(status: "ONLINE" | "BUSY" | "DO_NOT_DISTURB" | "BACK_IN_A_MOMENT"): void;
+```
+
+You can set the availability status of the current player. This status is visible to other players and can be used
+to indicate whether you're available for interaction.
+
+Supported statuses:
+- **"ONLINE"**: Clears any custom status (default state)
+- **"BUSY"**: Indicates the player is busy
+- **"DO_NOT_DISTURB"**: Indicates the player does not want to be disturbed
+- **"BACK_IN_A_MOMENT"**: Indicates the player is temporarily away
+
+```typescript
+// Set the player status to busy
+WA.player.setStatus("BUSY");
+
+// Clear the status (back to online)
+WA.player.setStatus("ONLINE");
+
+// Indicate you're temporarily away
+WA.player.setStatus("BACK_IN_A_MOMENT");
+```
+
 ## Detecting when the user enters/leaves a meeting
 
 ```ts


beingeek:
Hello @moufmouf  please review PR

play/src/front/Api/Events/SetStatusEvent.ts
```diff
@@ -0,0 +1,29 @@
...
+ * - BACK_IN_A_MOMENT: User is temporarily away
+ *
+ * Note: SILENT and AWAY are auto-managed by the system and cannot be set directly.
+ * Attempting to set them will result in a console warning.
+ */
+export const isSetStatusEvent = z.object({
+    status: z.union([
+        z.literal("ONLINE"),
+        z.literal("SILENT"), // Auto-managed, cannot be set directly
+        z.literal("AWAY"), // Auto-managed, cannot be set directly
```
moufmouf: Since these are auto-managed and cannot be set directly, we could simply remove those 2 lines.
```suggestion
```

play/src/front/Api/Iframe/player.ts
```diff
@@ -304,6 +304,30 @@ export class WorkadventurePlayerCommands extends IframeApiContribution<Workadven
...
+     *
+     * Supported statuses:
+     * - "ONLINE": Clear any custom status (default state)
+     * - "BUSY": Indicate the player is busy
+     * - "DO_NOT_DISTURB": Indicate the player does not want to be disturbed
+     * - "BACK_IN_A_MOMENT": Indicate the player is temporarily away
+     *
+     * Note: "SILENT" and "AWAY" are auto-managed by the system and cannot be set directly.
+     * Attempting to set them will result in a console warning with no effect.
+     *
```
moufmouf:
```suggestion
```

play/src/front/Api/IframeListener.ts
```diff
@@ -613,6 +616,31 @@ class IframeListener {
...
+                                BUSY: AvailabilityStatus.BUSY,
+                                DO_NOT_DISTURB: AvailabilityStatus.DO_NOT_DISTURB,
+                                BACK_IN_A_MOMENT: AvailabilityStatus.BACK_IN_A_MOMENT,
+                            };
+                            resetAllStatusStoreExcept(statusMap[statusValue]);
+                        } else {
+                            console.warn(
+                                `Status "${statusValue}" cannot be set via scripting API. Only ONLINE, BUSY, DO_NOT_DISTURB, and BACK_IN_A_MOMENT are allowed. SILENT and AWAY are auto-managed by the system.`
+                            );
+                        }
```
moufmouf: The code is correct, but we should move it somewhere else (in order to avoid `IframeListener` to grow too much.)

In the GameScene or in another class instanciated from the GameScene. In `IframeListener`, let's try to respect the global pattern (which is using RxJS observers to which we can subscribe)

moufmouf: Let's replace the "if" with a big switch case and a "default" with an ` const _exhaustiveCheck: never = ...`
This way, should we add a new status in the future, Typescript will raise an error if we don't handle it correctly.

beingeek: Resolved

beingeek: refactored

play/src/front/Api/Iframe/player.ts
beingeek: Resolved

play/src/front/Api/Events/SetStatusEvent.ts
beingeek: Removed

moufmouf:
Excellent! Thanks a lot!

PR #5867, opened 2026-03-17
Create page to install pwa

gparant:
https://github.com/user-attachments/assets/04f75251-989c-4fef-80e3-cd7127dc7d6e


play/src/front/Components/App.svelte
moufmouf, line 1: App.svelte is the Svelte root.
Here, we are polluting it with code specific for one feature (PẄA popup). We should probably move the <PwaInstallPrompt> into the GameOverlay and move the code from the script in a dedicated module? Maybe use a store for this?

Also, this is typically the kind of popup that should be implemented in the global onboarding system we talked about.

PR #5872, opened 2026-03-18
Create a notification to inform the duplicate user connected

gparant:
Screenshot 2026-03-18 at 11 10 48


play/src/front/Components/UI/DuplicateUserConnectedModal.svelte
```diff
@@ -0,0 +1,52 @@
...
+    } from "../../Stores/DuplicateUserConnectedStore";
+    import { LL } from "../../../i18n/i18n-svelte";
+
+    let dontRemindAgain = false;
+
+    function confirmAndContinue() {
+        if (dontRemindAgain) {
+            try {
+                localStorage.setItem(DUPLICATE_USER_DONT_REMIND_KEY, "1");
+            } catch {
```
moufmouf, line 12: Don't write in localStorage directly. Use the localUserStore instead.

play/src/i18n/en-US/warning.ts
```diff
@@ -27,6 +27,13 @@ const warning: BaseTranslation = {
     backgroundProcessing: {
         failedToApply: "Failed to apply background effects",
     },
+    duplicateUserConnected: {
+        title: "Already connected",
+        message:
+            "This user is already connected to this room from another tab or device. To avoid conflicts, please close the other tab or window.",
```
moufmouf:
```suggestion
            "You are already connected to this room from another tab or device. To avoid conflicts, please close the other tab or window.",
```

PR #5890, opened 2026-03-24
Kick out all users of a map when it is deleted

angerenage:


moufmouf:
It seems the changes introduces a bug in E2E tests: tests/map_storage_upload.spec.ts:605:9 › Map-storage Upload API @nomobile › special characters support

PR #5903, opened 2026-03-26
fix: audio context video automatically mute / unmute

gparant:
The issue:
The video will muted if the user is connected to WorkAdventure and doesn't interact with the navigator. The navigator will not block you if you have PWA. Today, we are requiring the unmute video and force the displaying button to unmute.
Additionally, it is not possible for us to determine if the audio context has been suspended and if the video is still muted.

Update:
If there is no notification permission or PWA, and the user has suspended their audio context. The user status will automatically be set to "back in moment" and receive a notification to click.
Otherwise, we will never block the audio.


play/src/front/Components/Toasts/DoNotDisturbInfoToast.svelte
```diff
@@ -0,0 +1,49 @@
...
+            handleActivate();
+        }, 1000);
+        return () => clearInterval(interval);
+    });
+</script>
+
+<ToastContainer extraClasses="w-full min-w-72 max-w-sm sm:min-w-80 sm:max-w-md" theme="secondary" {toastUuid}>
+    <div class="flex flex-col gap-2">
+        <p class="m-0 text-sm leading-snug text-white text-center">
+            {$LL.chat.status.do_not_disturb()}. Cliquez pour redevenir disponible et discuter avec les autres.
```
moufmouf: We need to translate this message.

play/src/front/WebRtc/AudioContextManager.ts
```diff
@@ -94,6 +94,16 @@ class AudioContextManager {
...
         return this.audioContexts.size;
     }
+
+    // Create a new context and verify that the context is not suspended.
+    // This context will not saved and just used to verify that the context is not suspended for the current page.
+    public verifyContextIsNotSuspended(): boolean {
+        // Chrome documentation : https://developer.chrome.com/blog/autoplay
+        // API : https://developer.mozilla.org/fr/docs/Web/API/AudioContext
+        // We need to create a new AudioContext and verify that the context is not suspended.
+        const context = new AudioContext();
```
moufmouf, line 109: I think I remember there is a recommendation not to build too many audio contexts.

You could get the default one from this class instead.

```suggestion
        const context = this.getContext();
```

play/src/front/Phaser/Game/GameScene.ts
```diff
@@ -2333,6 +2337,25 @@ export class GameScene extends DirtyScene {
...
+
+                    // Test if the user is in a PWA
+                    const isPWAInstalled = (navigator as Navigator & { standalone?: boolean }).standalone ?? false;
+
+                    // If the user has not allowed play sound, no notification permission and is not in a PWA, we need to show a message to the user to allow the page to play audio.
+                    if (!hasNotification && !isPWAInstalled) {
+                        console.warn("Audio context is suspended. Please allow the page to play audio.");
+                        // Update the user status. This is a specific status to indicate that the user has not allow to receive audio.
+                        // Set the proximity meeting to false and have the DENY_PROXIMITY_MEETING status.
+                        requestedStatusStore.set(AvailabilityStatus.BACK_IN_A_MOMENT);
```
moufmouf: We should not set the status to BACK_IN_A_MOMENT if the status is already set by the user to something restricted (like DO_NOT_DISTURB)

play/src/front/Components/Toasts/DoNotDisturbInfoToast.svelte
```diff
@@ -0,0 +1,49 @@
...
+        </p>
```
moufmouf, line 66: To be discussed: I think I would rather display a message like:

"Back in a moment -- click anywhere to switch online"
or "Your sound is muted -- click anywhere to switch online"

+ I would add a listener on the status of the audioManager that automatically disable this popup whenever the audioContext becomes available in the app (i.e. as soon as the user interacts with the page)

gparant: I added comment on that, the "changestatus" not works on navigator 😭
So I check every second if is active and close it

play/src/front/Phaser/Game/GameScene.ts
gparant: When we start WorkAdventure, the status can't be "DO_NOT_DISTURB"

play/src/front/WebRtc/AudioContextManager.ts
gparant: Not working because the function to get context save and return old context that was not updated with the new context of the navigator and the status was not changed.
So we stayed in "suspend" and could not close the "DoNotDisturbInfoToast"

PR #5911, opened 2026-03-27
refactor(play): centralize network viewport pipeline and align outbound paths

ydiarra:


play/src/front/Phaser/Game/GameScene.ts
```diff
@@ -1478,106 +1517,99 @@ export class GameScene extends DirtyScene {
...
         }
+        return Array.from(gameMapAreas.getAreas().values(), (area) => ({
+            x: area.x,
+            y: area.y,
+            width: area.width,
+            height: area.height,
+            hasMaxUsersInAreaProperty: area.properties.some(
+                (property) => property.type === "maxUsersInAreaPropertyData"
+            ),
+        }));
```
moufmouf: We don't need to extend the viewport to nearby areas anymore, because now, the back server is in charge of locked and max users areas.
So we can switch back to something way more simple, that only tracks the real viewport.
No need for MAX_AREA_EXTENSION_DISTANCE or MAX_VIEWPORT_SIZE anymore.

play/src/front/Connection/RoomConnection.ts
```diff
@@ -137,6 +137,19 @@ import { ConnectionClosedError } from "./ConnectionClosedError";
...
+function logViewportWorldCoordinatesSent(context: string, viewport: ViewportInterface): void {
+    viewportOutboundDebug(
+        "%s left=%d top=%d right=%d bottom=%d (world px, floored as on wire)",
+        context,
+        Math.floor(viewport.left),
+        Math.floor(viewport.top),
+        Math.floor(viewport.right),
+        Math.floor(viewport.bottom)
+    );
+}
```
moufmouf: Let's maybe remove this. Even in debug, it's going to be way too verbose.

PR #5923, opened 2026-03-31
Matrix connection for guest user

gparant:
TODO
 - Have a button to propose at the user to be connected as guest
 - Create option for admin to avoid guest connection


ydiarra:
I don't think we need to go through the pusher here. We can probably use the Matrix JS SDK registerGuest flow directly in the frontend to create the guest account. That would be simpler, and it is also closer to how we used to handle guest accounts before they were removed in [#4221](https://github.com/workadventure/workadventure/pull/4221)

moufmouf:
Wait, we need to discuss this feature before going further.

@gparant We should really consider if we want to use the `allow_guest_access` feature of Matrix first before moving forward.

As I remember for the latest Matrix meeting I had (in early 2025), I think literally noone on the Element side is using the `allow_guest_access` feature. I'm not even sure this is properly supported with the upcoming new OIDC login system for Matrix (MAS, etc...). So we might hit a lot of corner cases.

From a DB admin standpoint, this can be a problem. If any anonymous user gets a new Matrix account, the Matrix database will grow very much faster than it is now, with users that we don't care about. So we would need this PR to properly delete guest users and messages after some time.

Also, from a user experience, this can be misleading. If a guest comes to you and send you a message, then leaves and never comes back, you can still reply to that user (and the user will never get the reply, because we don't have the user email to send the reply).

There are really 2 use cases here:

1. Being able to send text messages to one user when we are anonymous
2. Being able to participate in a chat room when we are anonymous

I'd say that the second point (Being able to participate in a chat room while anonymous) is not a good idea. We need to track who says what. If someone is sending racist comments in a public room, we need some accountability.

For the first point (Being able to send text messages to one user), we could use a simple space to relay the messages (like what we do in bubbles). And if the anonymous user is logging out, there will be no-one in that space anymore and we could prevent the user at the other end from writing in the chat (making it clear no-one will receive the message).

gparant:
> Wait, we need to discuss this feature before going further.
> 
> @gparant We should really consider if we want to use the `allow_guest_access` feature of Matrix first before moving forward.
> 
> As I remember for the latest Matrix meeting I had (in early 2025), I think literally noone on the Element side is using the `allow_guest_access` feature. I'm not even sure this is properly supported with the upcoming new OIDC login system for Matrix (MAS, etc...). So we might hit a lot of corner cases.
> 
> From a DB admin standpoint, this can be a problem. If any anonymous user gets a new Matrix account, the Matrix database will grow very much faster than it is now, with users that we don't care about. So we would need this PR to properly delete guest users and messages after some time.
> 
> Also, from a user experience, this can be misleading. If a guest comes to you and send you a message, then leaves and never comes back, you can still reply to that user (and the user will never get the reply, because we don't have the user email to send the reply).
> 
> There are really 2 use cases here:
> 
> 1. Being able to send text messages to one user when we are anonymous
> 2. Being able to participate in a chat room when we are anonymous
> 
> I'd say that the second point (Being able to participate in a chat room while anonymous) is not a good idea. We need to track who says what. If someone is sending racist comments in a public room, we need some accountability.
> 
> For the first point (Being able to send text messages to one user), we could use a simple space to relay the messages (like what we do in bubbles). And if the anonymous user is logging out, there will be no-one in that space anymore and we could prevent the user at the other end from writing in the chat (making it clear no-one will receive the message).

I think that the use case is that I can be connected to Matrix without having to connect with my email.
In the opened world, that anyone can meet with you, exchange, and be connected on Matrix or not, who says what will not change based on the track. Whether it's Matrix or not, the guest won't be tracked tomorrow.

My opinion is that in terms of user experience, it is not acceptable to send a message and say that it will be not received. The guest user can use the same navigator to connect to Matrix and return without losing any data or information. I think that we need to go on this goal-oriented experience.

One use case is that if you are not connected to Matrix, during a meeting with Matrix live channel, the user connected cannot participate in the discussion. Same thing regarding the listener zone. The user experience is not perfect and we need to make improvements.

Finally, I think that if we have a service to manage message Matrix, we need to include all of the discussion chat features in the scope of the service. It will be simpler to comprehend and maintain throughout the code.

PR #5942, opened 2026-04-04
fix(i18n): improve Japanese translation for DO_NOT_DISTURB status label

Arata1202:
## Summary

Improve the Japanese translation for the `DO_NOT_DISTURB` status label in `play/src/i18n/ja-JP/actionbar.ts`.

## Change

| Key | Before | After |
|-----|--------|-------|
| `status.DO_NOT_DISTURB` | 邪魔しないで | 応答不可 |

## Reason

"邪魔しないで" sounds rude and unnatural in Japanese.


moufmouf:
Thanks! :+1:

PR #5965, opened 2026-04-13
Add ImageCoWebsite for image preview

angerenage:


play/src/front/Components/Cowebsites/ImageCowebsiteComponent.svelte
```diff
@@ -0,0 +1,186 @@
...
+                  padding: 48,
+                  maxZoom: 12,
+                  friction: 0.9,
+                  render,
+              } satisfies PanZoomOptions);
+
+    $: if (imageUrl && imageUrl !== loadedUrl) {
+        loadedUrl = imageUrl;
+        // eslint-disable-next-line @typescript-eslint/no-floating-promises
+        loadImage(imageUrl);
```
moufmouf: You can put a ".catch" here to remove the eslint warning.

And "loadImage" should reject on error (with the error object) instead of returning false (whose value is not taken into account)

PR #5974, opened 2026-04-14
Change username rendering to use a seperate canvas and adapt scale

### LINKED ISSUE #5893: Bad performance of WA.player.setOutlineColor
Feedback received:

> We found the reason for the „lags“.
> We used the outline color to highlight logged in users. This will degrade the performance dramatically.
> 
> WA.player.setOutlineColor(68, 153, 169);
> 
> After commenting out the line, the performance is fine.

We should try to reproduce this in a performance load test.

The outline is computed using a shader. Maybe we should see if we can preprocess the shader once and integrate it in the sprite that contains the user name.

angerenage:
The user name will now be displayed bigger when you zoom out (we try to keep it readable even when you zoom out a lot)

Furthermore, the username rendering is now sent to a sprite, which fixes a performance issue when a lot of names were displayed with outline.

Closes #5893

### DIFF
--- a/play/src/front/Phaser/Entity/Character.ts
+++ b/play/src/front/Phaser/Entity/Character.ts
@@ -16,7 +16,6 @@ import { CharacterTextureError } from "../../Exception/CharacterTextureError";
 import { getPlayerAnimations, PlayerAnimationTypes } from "../Player/Animation";
 import { ProtobufClientUtils } from "../../Network/ProtobufClientUtils";
 import { SpeakerIcon } from "../Components/SpeakerIcon";
-import { waScaleManager } from "../Services/WaScaleManager";
 
 import { UsernameDisplay } from "../Components/UsernameDisplay";
 import { lazyLoadPlayerCharacterTextures } from "./PlayerTexturesLoadingManager";
@@ -164,7 +163,6 @@ export abstract class Character extends Container implements OutlineableInterfac
 
             const playerNameOutlineColor = get(this.outlineColorStore);
             this.usernameDisplay = new UsernameDisplay(scene, 0, playerNameY, this.playerName, playerNameOutlineColor);
-            this.usernameDisplay.setScale(this.getSpriteScale(waScaleManager.zoomModifier));
             this.usernameDisplay.setAvailabilityStatus(this.availabilityStatus, true, true);
             this.add(this.usernameDisplay);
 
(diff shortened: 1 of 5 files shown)


play/src/front/Phaser/Entity/Character.ts
```diff
@@ -333,16 +318,23 @@ export abstract class Character extends Container implements OutlineableInterfac
...
-        this.statusDot.setAvailabilityStatus(availabilityStatus, instant);
-        if (this.getAvailabilityStatus() === AvailabilityStatus.SPEAKER) {
-            this.megaphoneIcon.show(true, false);
-        } else {
-            this.megaphoneIcon.show(false, false);
+        if (availabilityStatus !== AvailabilityStatus.UNCHANGED) {
+            this.availabilityStatus = availabilityStatus;
         }
+        this.usernameDisplay?.setAvailabilityStatus(availabilityStatus, instant);
+        this.usernameDisplay?.showMegaphone(this.availabilityStatus === AvailabilityStatus.SPEAKER, false);
```
moufmouf: We could use the UserNameDisplay.setAvailabilityStatus method to also decide if the megaphone is shown or not (and therefore, we could turn showMegaphone as a private method).

play/src/front/Phaser/Entity/Character.ts
```diff
@@ -570,22 +561,70 @@ export abstract class Character extends Container implements OutlineableInterfac
...
+        const previousTextureKey = this.playerNameTextureKey;
+        this.playerNameTextureKey = nextTextureKey;
+        this.usernameDisplay.setPlayerNameTexture(nextTextureKey);
+
+        if (previousTextureKey && this.scene.textures.exists(previousTextureKey)) {
+            this.scene.textures.remove(previousTextureKey);
+        }
+
+        this.usernameDisplay.setScale(this.getSpriteScale(waScaleManager.zoomModifier));
+        this.scene.markDirty();
```
moufmouf: Those 2 methods probably belong to the new UsernameDisplay class.

play/src/front/Phaser/Game/GameScene.ts
moufmouf, line 1: Instead of handling the subscription to CameraManagerEvent.CameraUpdate in the `GameScene`, we could subscribe to `CameraManagerEvent.CameraUpdate` in each `Character`  instances (directly in `Character.ts`, using: `this.gameScene.getCameraManager().on(...)`.

This would remove one responsibility from the GameScene. What do you think of this idea?

PR #5983, opened 2026-04-15
feat: use space queries for start and stop recording instead of metadata

ydiarra:


back/src/Model/MetadataProcessorInit.ts
```diff
@@ -1,55 +0,0 @@
...
-    const recordingState = space.getRecordingState();
-
-    return {
-        recorder: recordingState.recorder,
-        recording: recordingState.isRecording,
-    };
-}
-
-metadataProcessor.registerMetadataProcessor("recording", async (value, senderId, space) => {
-    const recordingMetadata = recordingMetadataSchema.safeParse(value);
```
moufmouf, line 23: We could keep this to forbid users to modify the "recording" metadata themselves, since it's now handled by the server.

play/src/front/Components/ActionBar/MenuIcons/RecordingMenuItem.svelte
```diff
@@ -104,11 +109,15 @@
 
 <ActionBarButton
     on:click={() => {
-        requestRecording();
+        requestRecording().catch((error) => {
+            console.error(`Failed to request recording`, error);
```
moufmouf, line 113: Maybe we could switch the button state to something that makes it explicit something went wrong here?
Like a danger state?

PR #5998, opened 2026-04-20
Create class to save query for members

gparant:


play/src/front/Chat/UserProvider/AdminChatMembersBrowserCache.ts
moufmouf, line 1: Isn't it a bit overkill? Maybe just in memory could be enough? And this way, if people don't see the member they expect to see, just refreshing will do the trick?

moufmouf:
I took the liberty to implement a simpler approach in #6000

PR #6012, opened 2026-04-22
fix: Matrix rooms could show up under **Group conversations**

gparant:
## Context

Matrix rooms could show up under **Group conversations** (the flat root list) right after an **invite** or **join** event, even when the room is attached to a parent space on the homeserver (`m.space.child` on the parent, `m.space.parent` on the child).

## What changed

- **On `RoomEvent.MyMembership` for `invite` and `join`**: before calling `manageRoomOrFolder`, we now **refresh joined space folders recursively** by calling `getChildren()` on each joined `MatrixRoomFolder`, so local state is rebuilt from `m.space.child` and invited/joined children appear under the correct folder when the parent space is already known to the client.
- **`handleOrphanRoom`**: if the room id is **already under** some folder in the `roomFolders` tree, we **skip** adding it to the root list, avoiding duplicates when the refresh has already placed the room under a folder but `manageRoomOrFolder` would still treat it as “orphan” for a moment.

## Cleanup

- Removed debug **`console.log` / `console.trace`** from `MatrixChatConnection`, `MatrixChatRoom`, and `MatrixRoomFolder` (kept **`console.error`** / **Sentry** for real failures).

## How to verify

- Receive an **invite** to a child room of a space you are **already joined** to: it should list under that space folder, not only under Group conversations.
- **Join** a child room of a joined space: same expectation.
- Edge case unchanged by design: invite/join to a child room when the **parent space is not in the client** (not a member / not synced) may still not show under that folder until the parent is visible to the SDK.

### DIFF
--- a/play/src/front/Chat/Connection/Matrix/MatrixChatConnection.ts
+++ b/play/src/front/Chat/Connection/Matrix/MatrixChatConnection.ts
@@ -203,7 +203,7 @@ export class MatrixChatConnection implements ChatConnectionInterface, MatrixChat
 
         this.folders = derived(
             [this.roomFolders, ...Array.from(this.roomFolders.values()).map((folder) => folder.myMembership)],
-            (folderList) => {
+            () => {
                 return Array.from(this.roomFolders.values()).filter(
                     (folder) => get(folder.myMembership) === KnownMembership.Join
                 );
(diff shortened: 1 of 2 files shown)


play/src/front/Chat/Connection/Matrix/MatrixChatConnection.ts
```diff
@@ -792,12 +791,58 @@ export class MatrixChatConnection implements ChatConnectionInterface, MatrixChat
...
 
+    /**
+     * Re-syncs each joined space folder from local `m.space.child` state so invited/joined
+     * children appear under the right folder before {@link #manageRoomOrFolder} runs.
+     */
+    private refreshJoinedFolderSubtree(folder: MatrixRoomFolder): void {
+        if (get(folder.myMembership) !== KnownMembership.Join) {
+            return;
+        }
+        folder.getChildren();
```
ydiarra, line 802: getChildren() may be called multiple times , but it currently recreates entries in roomList / folderList each time without cleaning up the previous instances.

We could instead reuse existing nodes when they are already present: keep the existing instance for regular rooms, and for folders reuse the existing folder and call getChildren() on it to refresh nested rooms.

PR #6024, opened 2026-04-23
fix: update get children to improve matrix performances

gparant:


play/src/front/Chat/Connection/Matrix/MatrixRoomFolder.ts
```diff
@@ -361,12 +361,28 @@ export class MatrixRoomFolder extends MatrixChatRoom implements RoomFolder {
                 );
             });
 
+        const childIds = new Set(children.map((c) => c.roomId));
+
+        for (const id of Array.from(this.roomList.keys())) {
```
ydiarra, line 380: Before deleting rooms or folders from the lists, we should call destroy() on the existing instance to properly clean up its subscriptions/listeners.

PR #6026, opened 2026-04-23
fix: chat connection perf updated to sync all Matrix data

gparant:


ydiarra (approved):
Looks good to me overall. I’m just wondering whether this flag ends up being equivalent to matrixClient.isInitialSyncComplete(), but I’m fine merging it as is

gparant:
> isInitialSyncComplete

Oh, may be could be better, let me check that 💪

PR #6028, opened 2026-04-23
docs(env): add WAM URL format to START_ROOM_URL comment

spiccinini:
It took me a while to understand that the map storage use the `/~/` prefix. This change may help others too.


moufmouf:
Excellent idea!

PR #6029, opened 2026-04-24
Implement a reconnection mechanism for dirty front disconnection

angerenage:


play/src/pusher/controllers/IoSocketController.ts
```diff
@@ -221,496 +219,373 @@ export class IoSocketController {
...
+                companionTextureId: z.string().optional(),
+                lastCommandId: z.string().optional(),
+                version: z.string(),
+                chatID: z.string(),
+                roomName: z.string(),
+                cameraState: z.string().transform((val) => val === "true"),
+                microphoneState: z.string().transform((val) => val === "true"),
+                tabId: z.string(),
+            }),
+            upgrade: async ({ query, request, isAborted, upgrade }) => {
```
moufmouf:
```suggestion
            upgrade: async ({ query, request, isAborted, upgrade, reject }) => {
```

We can add a "reject" method that does an upgrade with the UpgradeFailedData type (and "upgrade" only takes a ConnectingSocketData)

play/src/pusher/models/Websocket/SocketData.ts
```diff
@@ -29,7 +30,7 @@ export type SpaceName = string;
  */
 export type ConnectingSocketData = {
     rejected: false;
-    disconnecting: boolean;
+    disconnecting: boolean; // TODO: move the connection status management inside the PusherRoomSocketController
```
moufmouf: Indeed

play/src/front/Connection/WorkAdventureWebSocket.ts
moufmouf, line 1: Rename this file in WorkAdventureWebSocket.ts

play/src/front/Connection/WebSocket.ts
```diff
@@ -0,0 +1,187 @@
...
+
+        for (const { payload } of storedMessages) {
+            this.socket.send(payload);
+        }
+    }
+
+    private shouldReconnect(event: CloseEvent): boolean {
+        if (this.reconnectAttempted) return false;
+        // Do not reconnect if handshake/auth was refused.
+        if (event.code === 1008 || event.code === 4001 || event.code === 4003) return false;
```
moufmouf:
```suggestion
        if (event.code === 1008) return false;
```

play/src/front/Connection/WebSocket.ts
```diff
@@ -0,0 +1,187 @@
...
+            reason: event.reason,
+            wasClean: event.wasClean,
+        });
+        this.onclose?.call(this, closeEvent);
+    };
+
+    private handleErrorEvent = (): void => {
+        this.pendingErrorEvent = new Event("error");
+    };
+
```
moufmouf: Here, you swallow completely the error that triggered this.


```suggestion
    private handleErrorEvent = (event: Event): void => {
        this.pendingErrorEvent = event;
    };

```

play/src/front/Connection/WebSocket.ts
```diff
@@ -0,0 +1,187 @@
...
+    private flushPendingErrorEvent(): void {
```
moufmouf: This looks suspicious. flushPendingErrorEvent is only called on "close". If an error is triggered and "close" is never triggered (maybe because the connexion never opened), then the error event will be swallowed.

play/src/front/Connection/WebSocket.ts
```diff
@@ -0,0 +1,187 @@
...
+        this.onopen?.call(this, event);
+    };
+
+    private handleCloseEvent = (event: CloseEvent): void => {
+        this.detachSocketListeners(this.socket);
+
+        if (!this.manuallyClosed && this.shouldReconnect(event)) {
+            this.reconnectAttempted = true;
+            this.pendingErrorEvent = undefined;
+            this.reconnect();
```
moufmouf: Here, we should add a wait step (maybe not the first retry, but if we have many retries, let's wait 1 second, up to X seconds, X being configurable)

play/src/front/Connection/WorkAdventureWebSocket.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    public static setWebsocketFactory(websocketFactory: WebSocketFactory | null): void {
+        WorkAdventureWebSocket.websocketFactory = websocketFactory;
+    }
+
+    public onopen: ((this: WorkAdventureWebSocket, ev: Event) => void) | null = null;
+    public onclose: ((this: WorkAdventureWebSocket, ev: CloseEvent) => void) | null = null;
+    public onerror: ((this: WorkAdventureWebSocket, ev: Event) => void) | null = null;
+    public onmessage: ((this: WorkAdventureWebSocket, ev: MessageEvent<ServerToClientMessage>) => void) | null = null;
+
```
moufmouf, line 24: We should have some events to listen to when a reconnection process is starting and when a reconnection process succeeded.

moufmouf (commented):
We need to configure the reverse proxy to try targetting the same pusher each time (both in docker-compose.prod.yaml and in the Helm chart).

In the Helm chart, try adding annotations for both Nginx controller and Traefik.

TODO: testing with a Wifi connection cut.

moufmouf (commented):
We need to handle sticky sessions in Traefik in docker-compose.prod.yaml and in both Nginx/Traefik in K8S.

PR #6056, opened 2026-05-04
Replace player name text with DOM element

angerenage:


play/src/front/Phaser/Components/UsernameDisplay.ts
moufmouf, line 1: Let's keep the font to the same size now (no need for bigger fonts for non european characters, you can remove  

 this.playerNameFontSize = StringUtils.containsNonLatinCharacters(playerName) ? 11 : 8;

)

PR #6062, opened 2026-05-05
feat : added allow talking and forming bubbles in audience zone - by …

Harikesh04:
Adds an **"Allow talking and forming bubbles"** toggle to the Listener Megaphone area property in the map editor.

---

### Changes

* Added `allowTalking` boolean field (default: `false`) to `ListenerMegaphonePropertyData` schema
* Added a toggle switch in `ListenerMegaphonePropertyEditor.svelte` for the new property
* Updated `AreasPropertiesListener.ts` to conditionally set `isListenerStore`:

  * When `allowTalking` is `true`, users in the audience zone can still talk and form proximity chat bubbles instead of being muted as listeners
* Added i18n translation key for the new label

---

### Behavior

* **Default (`false`)**:
  Audience zone works as before — users are set as listeners (muted)

* **Enabled (`true`)**:
  Users in the audience zone can talk to each other and form bubbles while still receiving the megaphone broadcast
image
image

### DIFF
--- a/libs/map-editor/src/types.ts
+++ b/libs/map-editor/src/types.ts
@@ -143,6 +143,7 @@ export const ListenerMegaphonePropertyData = PropertyBase.extend({
     type: z.literal("listenerMegaphone"),
     speakerZoneName: z.string(),
     chatEnabled: z.boolean().default(false),
+    allowTalking: z.boolean().default(false),
     waitingLink: z.string().optional(),
 });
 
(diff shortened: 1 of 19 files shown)


moufmouf:
Note for me: before merging this, we probably need to review / modify the ActionMediaBox.svelte use of isListenerStore:

```
    <!-- Mute video -->
    {#if ($userIsAdminStore || !$isListenerStore) && !isScreenSharing}
        <button

```

This is used to decide if we display the "Mute video" button. Actually, we should decide that on a "per-space" basis as you could now be in a meeting AND in a listening zone. So the condition that decides if you can or cannot mute someone should probably be grounded in the psace or spaceUser object. For instance:

```
space.isModerator: boolean
```

The hard part: whether we are a moderator or not is hard to find from the space type alone (because some megaphone spaces are typed "ALL_STREAMS"). We will have to do a brainstorming to find the correct rule. @ydiarra

PR #6066, opened 2026-05-06
fix(play front): load fonts and boyscout

restray:
_Quick disclosure: I used AI **only** to generate this pull request summary. All the code was written by hand._

This PR fixes font loading in the Play frontend and improves style import structure.

## What changed

- Added explicit font imports (roboto, roboto-condensed, press-start-2p) in frontend styles.
- Introduced a dedicated play/src/front/style/font.scss file to centralize:
     - external font imports
     - custom @font-face declarations (Comic Jens, Pretendard Korean)
- Updated play/src/front/style/index.scss:
     - cleaned and reordered imports
     - switched to explicit .scss import paths
     - included font.scss in the main style entry
- Removed duplicated font-face declarations from style.scss.
- Removed redundant @fontsource/roboto dependency from play/package.json.
- Updated Tailwind shared style imports for fonts in libs/tailwind/style/tailwind.scss.
- Fixed Husky pre-commit shebang for better portability (#!/usr/bin/env bash).

## Why

- Ensure fonts (including boyscout/retro stack usage) are consistently loaded in Play.
- Avoid duplicated font definitions and improve style maintainability.
- Make script execution more reliable across environments.

## Impact

- Frontend styling/font rendering is more reliable and predictable.
- No expected functional behavior changes outside styling/font loading


moufmouf:
Salut @restray !

Ouah! Merci beaucoup!!! :heart:

I've been looking for this font loading issue for ages! (it was ok on Mac but not in Linux/Windows....) and you nailed it! Thanks a lot!

Ok, I need to ask because this is not obvious to me.... was the bug because of the duplicated font-face declarations? How did you spot that?

Also, I have a question: we created a Tailwind theme that the play container is relying on here: https://github.com/workadventure/design-system/

I'm not sure, but shouldn't we put the Roboto font faces in the theme instead of in play?

restray:
Hello @moufmouf,

I’m very new to this codebase, so I didn’t know you had a separate repository for the global styling, which was a bit surprising to me since some styling also seems to live in this repo.

The issue here was that Roboto / Roboto Condensed were not being loaded in the browser. What happened on your side is that your browser, relying on fonts already installed on your OS, already had access to Roboto and Roboto Condensed, which wouldn’t be the case for a typical non-developer user.

By the way, what’s the reason for having a separate repository for the styles?

Hope this message clears things up.

moufmouf:
I think we went with a separate repo for the styles because we are using the styles also for the website (https://workadventu.re) which is a Wordpress site (so completely unrelated to what we do with this repo :) )

Thanks for the explanation!

restray:
Oh, that could explain the missing font in the banner "Coming from Gather.town" ! I'll do a PR later today.

moufmouf:
Thanks a lot @restray !

PR #6079, opened 2026-05-11
fix(pusher): silent OIDC refresh token on access token expiry

BAWES:
## Problem

When a user's OIDC access token expires (Keycloak/Authentik default: ~5 min), the next `/me` call returns `401 Invalid token` and forces a full re-login, even though a valid refresh token was available.

**Root cause:** `refresh_token` from the OIDC `tokenSet` in `getUserInfo()` was silently discarded and never stored in the WA auth JWT.

---

### `OpenIDClient.ts`
- `getUserInfo()` return type now includes `refresh_token: string | undefined`
- New `refreshAccessToken(refreshToken)` method using `openid-client`'s built-in `client.refresh()` — no new dependencies

### `JWTTokenManager.ts`
- `AuthTokenData` schema gains optional `refreshToken` field
- `createAuthToken()` gains optional `refreshToken` parameter (all existing callers pass `undefined` transparently)

### `AuthenticateController.ts`
- `/openid-callback`: stores `refresh_token` in the WA JWT at first login
- `/me`: on `checkTokenAuth()` failure, attempts silent refresh before returning 401:
  1. Calls `refreshAccessToken()` to get fresh tokens
  2. Re-verifies new access token via `checkTokenAuth()`
  3. Returns a new WA JWT as `authToken` — client's existing `localUserStore.setAuthToken()` persists it with zero UI disruption
  4. If refresh also fails, falls through to the original `401`

---

## Behaviour matrix

| Scenario | Before | After |
|---|---|---|
| Access token expired, refresh token valid | 401, forced re-login | Silent exchange, no interruption |
| Refresh token also expired/revoked | 401, forced re-login | 401, forced re-login (unchanged) |
| Anonymous user (no accessToken) | unchanged | unchanged |
| Provider doesn't issue refresh tokens | 401, forced re-login | 401, forced re-login (graceful fallback) |

---

## Notes

- Works with all standard OIDC providers. Providers requiring `offline_access` scope to issue refresh tokens should add it to `OPID_SCOPE`.
- Refresh token rotation is handled — new `refresh_token` from provider replaces old one in the new WA JWT.
- No new env vars, no new dependencies, no frontend changes needed.

### DIFF
--- a/play/src/pusher/services/JWTTokenManager.ts
+++ b/play/src/pusher/services/JWTTokenManager.ts
@@ -1,19 +1,42 @@
-import Jwt from "jsonwebtoken";
+import { jwtVerify, SignJWT, errors } from "jose";
 import z from "zod";
 import { ADMIN_SOCKETS_TOKEN, SECRET_KEY } from "../enums/EnvironmentVariable";
 
 export const AuthTokenData = z.object({
     identifier: z.string(), //will be a email if logged in or an uuid if anonymous
     accessToken: z.string().optional(),
+    refreshToken: z.string().optional(),
     username: z.string().optional(),
     locale: z.string().optional(),
-    tags: z.string().array().optional(),
+    tags: z
+        .preprocess((val) => {
+            if (typeof val === "string") {
+                try {
+                    return JSON.parse(val);
+                } catch {
+                    return [val];
+                }
+            }
+            return val;
+        }, z.string().array())
+        .optional(),
     matrixUserId: z.string().optional(),
 });
 export type AuthTokenData = z.infer<typeof AuthTokenData>;
 
 export const AccessTokenData = z.object({
-    tags: z.string().array().optional(),
+    tags: z
+        .preprocess((val) => {
+            if (typeof val === "string") {
+                try {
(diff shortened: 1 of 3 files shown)


BAWES:
@moufmouf would love your input

play/src/pusher/services/JWTTokenManager.ts
moufmouf, line 21: I don't think this is related to the refresh token PR. I also think this was handled in some other PR already.

moufmouf, line 39: I don't think this is related to the refresh token PR. I also think this was handled in some other PR already.

moufmouf (commented):
Hey Khalid!

Excellent idea to add support for refresh tokens! I'm all in.

I think you started your PR from an old branch (or an old version of master) though.

Can you double check this? (I also asked a review by Copilot that can be useful)

BAWES:
Thanks will attempt another @moufmouf

BAWES:
Moved to https://github.com/workadventure/workadventure/pull/6085

PR #6085, opened 2026-05-12
fix(pusher): silent OIDC refresh token on access token expiry

BAWES:
## Problem

When a user's OIDC access token expires, the `/me` endpoint immediately throws a 401, forcing users to re-login even though a valid refresh token exists.

## Solution

Store the OIDC `refresh_token` inside the WorkAdventure JWT at login time, and silently exchange it for a new access token when `/me` detects an expired token — transparent to the user.

### `JWTTokenManager.ts`
- Added optional `refreshToken` field to `AuthTokenData` Zod schema
- Added `refreshToken` as optional parameter to `createAuthToken()`

### `OpenIDClient.ts`
- `getUserInfo()` now returns `refresh_token` from the OIDC token set
- Added `refreshAccessToken(refreshToken)` using `openid-client`'s built-in `client.refresh()` — no new dependencies

### `AuthenticateController.ts`
- `openIDCallback` passes `refresh_token` when minting the WA JWT
- `/me` catch block attempts silent token refresh before throwing 401; issues a re-minted `authToken` with rotated tokens on success

## No new dependencies

`openid-client` already exposes `client.refresh()` — nothing new to install.

## Backwards compatible

`refreshToken` is optional in the JWT schema. Existing tokens without it continue to work as before (no refresh attempted, falls through to 401).

### DIFF
--- a/play/src/pusher/controllers/AuthenticateController.ts
+++ b/play/src/pusher/controllers/AuthenticateController.ts
@@ -28,26 +28,34 @@ export class AuthenticateController extends BaseHttpController {
 
         let redirectToMatrixPath: string;
         if (fs.existsSync("dist/public/redirectToMatrix.html")) {
+            // In prod mode
             redirectToMatrixPath = "dist/public/redirectToMatrix.html";
         } else if (fs.existsSync("redirectToMatrix.html")) {
+            // In dev mode
             redirectToMatrixPath = "redirectToMatrix.html";
         } else {
             throw new Error("Could not find redirectToMatrix.html file");
         }
 
         this.redirectToMatrixFile = fs.readFileSync(redirectToMatrixPath, "utf8");
+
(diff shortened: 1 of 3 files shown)


BAWES:
@moufmouf This is the updated version

moufmouf:
Hi @BAWES !

Looks good to me. There is one thing that you would need to fix before we merge this: your PR is removing plenty of comments (some of whom are valuables). Can you keep your code as-is but keep the comments instead of removing them?

BAWES:
> Hi @BAWES !
> 
> 
> 
> Looks good to me. There is one thing that you would need to fix before we merge this: your PR is removing plenty of comments (some of whom are valuables). Can you keep your code as-is but keep the comments instead of removing them?

@moufmouf amazing! Does my latest commit resolve it?

PR #6153, opened 2026-05-27
feat: add push-to-talk temporary unmute

ydiarra:
## Problem

Users who are voluntarily muted in a WorkAdventure meeting need a quick way to speak briefly without toggling their microphone state permanently.

`Space` is the natural key for this (it is what Zoom uses), but it was already WorkAdventure's contextual **interact** key, so the two could not coexist: the shortcuts menu ended up listing `Space` twice with contradictory descriptions.

## Solution

Add a push-to-talk behavior on `Space` for WorkAdventure-controlled meetings:
- conversation bubbles
- LiveKit rooms

Holding `Space` temporarily enables the microphone, and releasing it returns the user to their previous muted state.

To free the key, **the interact shortcut moves from `Space` to `X`**. `E` and `F` — the usual interact keys in games — are already taken in WorkAdventure by the map editor and by "follow", so `X` was chosen as the closest free key to the movement cluster.

## ⚠️ Breaking change — action required by map creators

Texts that tell users to "press Space" are now wrong. WorkAdventure cannot rewrite them, because they are authored in the map:

- the `triggerMessage` property on entities and areas
- the `message` passed to `WA.ui.displayActionMessage()` and `WA.ui.displayPlayerMessage()`

Write the `[X]` marker in a trigger message and WorkAdventure renders it as a key badge.

## Changes

**Push-to-talk**
- Add a non-persistent temporary microphone state, separate from the persisted on/off choice; the effective microphone state is the OR of the two.
- Enable audio constraints while push-to-talk is active without persisting the microphone as unmuted. Privacy, energy-saving and availability-status blockers stay stronger than a temporary unmute.
- Release the temporary unmute on key release, window blur, hidden tab, or when push-to-talk stops being available, so the user can never get stuck live.
- Ignore push-to-talk while typing in text inputs, textareas, selects, or contenteditable elements.
- Update the microphone action bar button to show the active blue state while temporarily unmuted, and the local camera container so mute indicators and voice feedback reflect it.

**Interact key `Space` → `X`**
- `INTERACT_KEY` in `UserInputManager` is the single source of truth: the keyup case, the Phaser `keyup-<KEY>` event, the shortcuts table and the on-screen keycap all derive from it.
- `addSpaceEventListener` / `removeSpaceEventListener` become `addInteractEventListener` / `removeInteractEventListener` (interface, manager, and the 10 popup components).
- The interact case gains two guards a letter key needs and `Space` did not: the typing guard, and a Ctrl/Cmd bail-out so `Ctrl+X` stays "cut".
- The keycap SVG derives its label and width from `INTERACT_KEY` instead of hardcoding `SPACE` at a fixed width.
- `trigger.spaceKeyboard` becomes `trigger.interactKeyboard` = `[X]` across all 15 locales. Six locales spelled the key out in prose and never rendered a badge; they now use the marker too.
- Docs, JSDoc and the three E2E specs that pressed `Space` are updated.

**Bugs fixed on the way**
- Popups were never suppressed by push-to-talk: they listen on the Phaser `keyup-SPACE` emitter, a different channel from `handleKeyUpEvent`, so releasing `Space` after talking also fired the open popup. Gone now that the two features use different keys.
- Holding `Space` cancelled click-to-move: the dead `UserInputEvent.Interact` polled key set the event to true, and `Player.moveUser` drops `pathToFollow` on `anyExcept(SpeedUp)`. The entry is deleted — it had no consumer.
- The shortcuts menu showed `F , F`: "follow" was declared in both shortcut lists.

## Tests

- Unit coverage for the temporary microphone state, force-mute, the release watchdog, and the typing guard.
- The three E2E specs that used `Space` to interact now use `X`.


play/src/front/Stores/MicrophoneSessionStore.ts
```diff
@@ -0,0 +1,245 @@
...
+    stopTemporaryUnmute: () => void;
+    forceMuteMicrophone: () => void;
+}
+
+interface CreateMicrophoneSessionStoreOptions {
+    initialPersistentMicrophoneState: boolean;
+    persistPersistentMicrophoneState: (state: boolean) => void;
+}
+
+export function createMicrophoneSessionStore({
```
moufmouf: Maybe add some comments here to explain what those stores are useful for?

PR #6160, opened 2026-05-28
Adding a pulsing animation when speaking instead of the white bubble

moufmouf:


moufmouf:
We chose #6159 instead.

PR #6181, opened 2026-06-03
Support multi proximity chat rooms

ydiarra:

PR #6183, opened 2026-06-03
fix: enable talking and hearing within bubbles in audience zones

Harikesh04:
## Issue
  In an audience zone with "Allow talking and forming bubbles" ON, members could form a bubble but
  couldn't actually talk to or hear each other.
  
  ## Fix
  Talking audience members now stay regular proximity participants, so the backend forms a real bubble
  where they can talk and hear each other.

  ## Behavior
  | Allow talking | Result |
  |---|---|
  | ON | Form a bubble and talk/hear each other |
  | OFF | Silent listeners (unchanged) |

  ## Test
  - ON: members in the same bubble can talk and hear each other.
  - OFF: no bubble forms, members stay silent listeners.

// when Allow talking and forming bubbles is off

// when Allow talking and forming bubbles is on


ydiarra:
Thanks a lot for your PR and for taking the time to look into this issue.

I’m already working on a fix in another PR , The underlying problem is related to the way proximity chat rooms are currently handled: some parts of the code still assume a single proximity chat room/session, which can lead to inconsistent behavior when multiple proximity chat rooms are involved.

You can follow the progress on that other PR here: #6181 

Thanks again for your contribution!

PR #6185, opened 2026-06-04
feat(analytics): add generic analytics pipeline

ydiarra:
## Summary

Introduces a **generic analytics pipeline** that replaces the legacy `video-quality-batch` endpoint with a unified event bus shared by front, pusher and media sources. Events are buffered in the pusher, batched and sent to Admin, which stores them in ClickHouse (V1 contract).

Front-side, this adds:

- Meeting analytics as pusher-measured intervals (`meeting.started` / `meeting.ended`, opened and closed by the front through `timed_event.open` / `timed_event.close`, timed on the pusher's clock).
- Area & entity analytics in the map editor (create / update / delete commands now emit `mapeditor.*` events).
- Screenshare & cowebsite analytics (`screenshare.started/ended`, `cowebsite.opened` with media-kind detection).
- Session lifecycle events from `ConnectionManager` (`session.started/ended`).

Pusher-side, this adds:

- `AnalyticsEventsQueue` — bounded queue (default 10 000 events), 5 s flush interval, per-batch retry with exponential backoff, per-event split-retry on 422.
- `AnalyticsPresenceTracker` — tab-keyed presence map producing `user.connected` / `user.disconnected` events with computed durations.
- Per-world gating: pusher fetches `analyticsEventsEnabled` from Admin at JOIN_ROOM and drops everything for a world that opted out. Per-category filtering and anonymization are applied admin-side at ingestion (`AnalyticsEventsService::filterEvents`), which owns the policy — nothing is duplicated pusher-side.
- Video quality samples ride the generic pipeline as `media.video_quality.sample`. The legacy `video-quality-batch` queue, endpoint and capability are **removed** (see UPGRADE.md): Admin must be upgraded before or with the pusher, otherwise video quality analytics stop until it is.
- Graceful drain on SIGTERM / SIGINT so events buffered at shutdown reach Admin.
- Hardening: strict source whitelist (only `front` / `media` accepted from clients), 100-event cap per `analyticsEventReportMessage`, 8 KiB cap on per-event properties.

The event contract (`schemaVersion`, `sentAt`, `pusherInstanceId`, `events[]`) is documented in the admin-side companion change.

The matching admin work lives in the workadventure-saas repo, branch `generic-analytics-pipeline`.

## Why

Today only `video-quality-batch` is sent to admin; everything else goes to PostHog. That splits product analytics across two destinations, blocks per-world opt-out, and leaves self-hosted instances blind to product usage.

A single first-party pipeline with per-world / per-category opt-out unlocks:

- self-hosted product analytics without a third-party tool;
- per-world / per-category opt-out for compliance;
- a generic event catalogue (conversation, mapeditor, screenshare, cowebsite, …) usable for adoption, churn, onboarding and retention work in the SaaS admin.

## Test plan

- [ ] `npm run pretty-check` green
- [ ] `npm run typecheck` green
- [ ] `npm run test` (vitest) green — new suites: `ConversationAnalytics.test.ts`, `AnalyticsClient.test.ts`, `AnalyticsEventsQueue.test.ts`, `AnalyticsPresenceTracker.test.ts`
- [ ] e2e green (the previously failing Firefox 2/4 shards need re-confirming after rebase on master)
- [ ] Manual: start a spontaneous bubble then a meeting — confirm one `meeting.ended` row per interval, with `startedAt` / `endedAt` measured by the pusher, is stored admin-side
- [ ] Manual: set `analytics_events_enabled = false` on a world — confirm no event is stored for that world
- [ ] Manual: kill the pusher during an active session — confirm the last events still reach admin (SIGTERM drain)
- [ ] Manual: open a cowebsite with a sensitive query string (`?token=xxx`) — confirm the token does not appear in ClickHouse

## Notes for reviewers

- The event contract is stable for V1 (`schemaVersion: 1`). Any change implies bumping `schemaVersion` on both pusher and admin.
- The policy is captured once at JOIN_ROOM and **not refreshed** during the session — intentional for V1, to be revisited if needed.
- **Breaking:** the legacy `video-quality-batch` endpoint is retired, not kept as a fallback. Documented in UPGRADE.md together with the renamed `ANALYTICS_*` env vars.
- 422 from admin still triggers the per-event split-retry path so a single bad event does not poison a whole batch.

## Follow-ups (separate PRs)

- Optional on-disk buffer in front of the queue to survive multi-hour admin outages without loss
- Mid-session policy refresh (push admin → pusher via the existing capabilities channel)
- Stricter event-name allowlist on the pusher side (currently enforced admin-side)


libs/messages/src/JsonMessages/MapDetailsData.ts
```diff
@@ -106,6 +106,10 @@ const isLegalsData = z.object({
     description:
       "The link to the 'cookie policy' page (link displayed on the 'enter your name' scene)",
   }),
+  analyticsNoticeUrl: extendApi(z.string().nullable().optional(), {
+    description:
+      "The link to the analytics metrics notice page for this world.",
+  }),
```
moufmouf: To be removed

play/src/front/Administration/AnalyticsClient.ts
```diff
@@ -35,513 +86,1096 @@ class AnalyticsClient {
...
+                this.sendTimedEventReport,
+            ),
+        );
+    }
+
+    /**
+     * The caller no longer passes a duration, and cannot: the pusher measures the
+     * interval. This used to take a client-computed durationSeconds straight into the
+     * pipeline.
+     */
```
moufmouf:
```suggestion
```

play/src/front/Administration/TimedAnalyticsEvent.ts
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * These two frames are instructions, not events: they are intercepted in
+ * processAnalyticsReportMessage and never reach the queue. Grepping the admin for
+ * `timed_event.open` finds nothing, on purpose.
+ */
+export function openTimedAnalyticsEvent(
+    eventName: string,
+    properties: TimedEventProperties,
+    sendReport: (message: AnalyticsEventReportMessage) => void,
+): TimedAnalyticsEventHandle {
```
moufmouf: It would be great to have strong typing here (instead of a `Record<string, string | number | boolean | null | undefined>` for TimedEventProperties.
Ideally, openTimedAnalyticsEvent should be a function taking a generic

play/src/front/Administration/AnalyticsClient.ts
```diff
@@ -35,513 +86,1096 @@ class AnalyticsClient {
...
+            // from a dead socket, which the pusher drops as unpaired.
+            this.openAreas.clear();
+            this.openScreenShares.clear();
+            this.openStatus = undefined;
+            this.currentStatus = undefined;
+        }
+        this.flushPendingAdminEvents();
+    }
+
+    private trackAdminEvent(eventName: string, properties: AdminAnalyticsProperties = {}): void {
```
moufmouf: Could be a generic function.

play/src/i18n/it-IT/login.ts
```diff
@@ -15,6 +15,7 @@ const login: DeepPartial<Translation["login"]> = {
     termsOfUse: "termini di utilizzo",
     privacyPolicy: "politica sulla privacy",
     cookiePolicy: "politica sui cookie",
+    analyticsNotice: "informativa sull'analisi",
```
moufmouf: Translation to be removed

play/src/pusher/controllers/IoSocketController.ts
```diff
@@ -354,6 +356,8 @@ export class IoSocketController {
                         world: "",
                         chatID,
                         canRecord: false,
+                        analyticsEventsEnabled: true,
```
moufmouf:
```suggestion
                        analyticsEventsEnabled: false,
```

As long as there is no admin, there is no interest in getting Analytics from the client

play/src/pusher/controllers/IoSocketController.ts
```diff
@@ -1105,6 +1111,18 @@ export class IoSocketController {
                                     message.message.videoQualityReportMessage,
                                     socket.getUserData(),
                                 );
+                                analyticsEventsQueue.enqueueVideoQualityReport(
+                                    message.message.videoQualityReportMessage,
+                                    socket.getUserData(),
+                                );
```
moufmouf, line 1127: I think we should choose only one way to send events. Either via a generic analyticsEventsQueue (sounds good to me) or via the old videoQualityAnalyticsQueue. But we should not do both.

play/src/pusher/enums/EnvironmentVariableValidator.ts
```diff
@@ -283,16 +283,40 @@ export const EnvironmentVariables = z.object({
...
-        .describe("Interval in milliseconds between video quality analytics batch flushes. Defaults to 10000"),
+        .describe(
+            "Interval in milliseconds between analytics batch flushes. Also drives the generic analytics queue, which reuses the VIDEO_ANALYTICS_* settings instead of defining its own — tuning either queue tunes both. Defaults to 10000",
+        ),
     VIDEO_ANALYTICS_TIMEOUT_MS: PositiveIntAsString.optional()
         .transform((val) => toNumber(val, 2_000))
-        .describe("HTTP timeout in milliseconds for video quality analytics ingestion calls. Defaults to 2000"),
+        .describe(
+            "HTTP timeout in milliseconds for analytics ingestion calls. Also used by the generic analytics queue. Defaults to 2000",
+        ),
```
moufmouf, line 293: We should probably rename VIDEO_ANALYTICS_TIMEOUT_MS into ANALYTICS_TIMEOUT_MS

play/src/pusher/enums/EnvironmentVariableValidator.ts
```diff
@@ -283,16 +283,40 @@ export const EnvironmentVariables = z.object({
         .describe("Enable/disable the OpenAPI documentation endpoint. Defaults to false"),
     VIDEO_ANALYTICS_FLUSH_INTERVAL_MS: PositiveIntAsString.optional()
         .transform((val) => toNumber(val, 10_000))
-        .describe("Interval in milliseconds between video quality analytics batch flushes. Defaults to 10000"),
+        .describe(
+            "Interval in milliseconds between analytics batch flushes. Also drives the generic analytics queue, which reuses the VIDEO_ANALYTICS_* settings instead of defining its own — tuning either queue tunes both. Defaults to 10000",
```
moufmouf: Rename VIDEO_ANALYTICS_FLUSH_INTERVAL_MS to ANALYTICS_FLUSH_INTERVAL_MS

play/src/pusher/enums/EnvironmentVariableValidator.ts
```diff
@@ -283,16 +283,40 @@ export const EnvironmentVariables = z.object({
...
         .transform((val) => toNumber(val, 2_000))
-        .describe("HTTP timeout in milliseconds for video quality analytics ingestion calls. Defaults to 2000"),
+        .describe(
+            "HTTP timeout in milliseconds for analytics ingestion calls. Also used by the generic analytics queue. Defaults to 2000",
+        ),
     VIDEO_ANALYTICS_MAX_QUEUE_SIZE: PositiveIntAsString.optional()
         .transform((val) => toNumber(val, 10_000))
-        .describe("Maximum number of video quality samples queued in pusher memory. Defaults to 10000"),
+        .describe(
+            "Maximum number of analytics events queued in pusher memory. Caps the generic analytics queue as well as video quality samples. Defaults to 10000",
```
moufmouf: Rename VIDEO_ANALYTICS_MAX_QUEUE_SIZE to ANALYTICS_MAX_QUEUE_SIZE...

play/src/pusher/enums/EnvironmentVariableValidator.ts
```diff
@@ -283,16 +283,40 @@ export const EnvironmentVariables = z.object({
...
+        .transform((val) => toNumber(val, 20_000))
+        .describe(
+            "Maximum time in milliseconds spent draining the generic analytics queue on SIGTERM / SIGINT before the process exits. Must stay comfortably BELOW your orchestrator's grace period (Kubernetes terminationGracePeriodSeconds defaults to 30s), not merely equal to it: the drain has to finish and the process exit before SIGKILL lands, or the intervals it just closed die with it. Defaults to 20000, i.e. 10s of headroom under the Kubernetes default. Raising it above the grace period cannot buy more draining — it only converts a clean exit into a kill.",
+        ),
+    VIDEO_ANALYTICS_DRAIN_TIMEOUT_MS: PositiveIntAsString.optional()
+        .transform((val) => toNumber(val, 20_000))
+        .describe(
+            "Maximum time in milliseconds spent draining the legacy video-quality analytics queue on SIGTERM / SIGINT before the process exits. Both queues drain in parallel and the process waits for both, so the shutdown budget is the LARGER of the two: raising this one alone is enough to push the exit past the grace period. Defaults to 20000",
+        ),
```
moufmouf: Merge VIDEO_ANALYTICS_DRAIN_TIMEOUT_MS and ANALYTICS_DRAIN_TIMEOUT_MS into a single DRAIN_TIMEOUT_MS that can be used for any services waiting for shutdown.

play/src/pusher/models/Websocket/SocketData.ts
```diff
@@ -22,6 +22,19 @@ export interface BackSpaceConnection extends BackSpaceConnection_ {
...
+    | "collaboration_activity"
+    | "workspace_actions"
+    | "quality_diagnostics"
+    | "user_level_activity";
+
+export type AnalyticsMetricsPolicy = {
+    schemaVersion: number;
+    legalTemplateVersion: string;
+    categories: Partial<Record<AnalyticsMetricCategory, boolean>>;
+};
```
moufmouf: To be removed

play/src/pusher/services/AdminApi.ts
```diff
@@ -55,6 +55,18 @@ export const AdminLoginMessage = z.object({
...
+    legalTemplateVersion: z.string(),
+    categories: z.object({
+        presence_sessions: z.boolean().optional(),
+        collaboration_activity: z.boolean().optional(),
+        workspace_actions: z.boolean().optional(),
+        quality_diagnostics: z.boolean().optional(),
+        user_level_activity: z.boolean().optional(),
+    }),
+});
+
```
moufmouf: To be removed

play/src/pusher/services/AdminApi.ts
```diff
@@ -125,6 +137,12 @@ export const isFetchMemberDataByUuidSuccessResponse = z.object({
         description:
             "True if the user can record the room. In addition to this, the user still needs to have the correct tags as defined in the WAM settings.",
     }),
+    analyticsEventsEnabled: extendApi(z.boolean().optional(), {
+        description: "True if admin analytics events can be sent for this world.",
+    }),
+    analyticsMetricsPolicy: extendApi(isAnalyticsMetricsPolicy.optional(), {
+        description: "Versioned category policy controlling which admin analytics metrics can be sent for this world.",
+    }),
```
moufmouf: To be removed

play/src/pusher/services/AnalyticsEventCatalog.ts
```diff
@@ -0,0 +1,1079 @@
...
+);
+
+/* -------------------------------------------------------------------------- */
+/*                        Auth / session / room                               */
+/* -------------------------------------------------------------------------- */
+
+export const authUserIdentifiedEvent = defineEvent(
+    "auth.user_identified",
+    z.object({ roomId: z.string().nullable().describe("Room the user was identified in, when known.") }),
+    "The user was identified, i.e. attached to an account rather than anonymous.",
```
moufmouf: Let's try to build one big Zod object in common instead of pusher with this structure:

```
z.union([
    z.object({
        type: "auth.user_identified",
        parameters: z.object({ roomId: z.string().nullable().describe("Room the user was identified in, when known.") }),
    }),
    // ...
]);
```

moufmouf: Usage of this Zod object:

- in the front (to gain type safety)
- in the pusher Swagger controller
  - at SwaggerGenerator.ts, register the object in "SwaggerGenerator.definitions"
  - add a "@openapi" section (just like in AdminApi.ts) in `AnalyticsEventQueue.postBatch`

play/src/pusher/services/AnalyticsEventSchema.ts
```diff
@@ -0,0 +1,92 @@
...
+    // Stated by the client when it closes its own interval. `type_changed` is the
+    // one consumers cannot do without: it means one conversation was split because
+    // its type changed, so a bubble that became a meeting reports two intervals and
+    // has to be stitched on time rather than by id.
+    "closed_by_client",
+    "left_conversation",
+    "left_area",
+    "type_changed",
+    "status_changed",
+    "cleanup",
```
moufmouf: Probably not very useful. We can remove the reason why an event stops on the client side and simply say it's the client that triggered the end.

play/src/pusher/services/AnalyticsEventsQueue.ts
```diff
@@ -0,0 +1,814 @@
...
+    return {
+        ...event,
+        userUuid: anonymousIdentifier(event.world, event.userUuid, secret),
+        userId: null,
+        spaceUserId: anonymousIdentifier(event.world, event.spaceUserId, secret),
+        clientIp: null,
+        tabId: null,
+        properties: anonymizeProperties(event.properties),
+    };
+}
```
moufmouf: Let's completely push the notion of anonymization into the admin. Let's make the pusher and the front 100% unaware of those. The pusher shall send all information, always, to the admin and the admin will have the job to filter anonymous data if the client requires so.
Having the admin own this part will simplify the code here and remove a whole category of bugs.

play/src/pusher/services/AnalyticsPresenceTracker.ts
moufmouf, line 1: We can probably replace the AnalyticsPresenceTracker with the more generic `analyticsTimedEventTracker` ?

play/src/pusher/services/AnalyticsReportMessageHandler.ts
```diff
@@ -0,0 +1,187 @@
...
+ * `connectedAt` / `disconnectedAt` / `durationSeconds` straight from the event
+ * properties. A client sending one of these under source "front" would therefore
+ * write session rows of its own choosing and inflate the world's connection time
+ * and risk score. `media.video_quality.sample` is likewise synthesized from
+ * videoQualityReportMessage, not reported directly.
+ *
+ * The `media.` prefix as such stays open — the front legitimately emits other
+ * `media.*` events; only this exact name is reserved.
+ */
+const PUSHER_SYNTHESIZED_EVENT_NAMES = ["user.connected", "user.disconnected", "media.video_quality.sample"];
```
moufmouf: This will become useless when we Zod the allowed events from the front. We will have a white-listed value of allowed events, so no need to protect against invalid value another way than Zod.

play/src/pusher/services/AnalyticsTimedEventTracker.ts
```diff
@@ -0,0 +1,234 @@
...
+ * same `play` service and deploy together, so the only skew this can cause is a tab
+ * from a new replica reconnecting to an old one mid-deploy; the degraded outcome is
+ * a rejected open, i.e. what an unknown name already does today.
+ */
+export const TIMED_EVENT_NAMES = new Set<string>([
+    "conversation.ended",
+    "area.dwell",
+    "meeting.screenshare.ended",
+    "status.dwell",
+]);
```
moufmouf: Probably useless if we manage to have a proper Zod type to validate client stop events

play/src/server.ts
```diff
@@ -65,6 +71,88 @@ if (SENTRY_DSN != undefined) {
...
+                console.error(`Error while draining the ${drains[index][0]} queue during shutdown`, result.reason);
+            }
+        });
+    })()
+        .catch((error) => {
+            console.error("Error while draining analytics queues during shutdown", error);
+        })
+        .finally(() => {
+            process.exit(exitCode);
+        });
```
moufmouf: It would be great to have less code related to drains in this file.
Maybe by having a way for services to register their drains?

play/src/pusher/services/AnalyticsEventsQueue.ts
```diff
@@ -0,0 +1,719 @@
...
+                clientEventTimeMs: event.clientEventTimeMs,
+            });
+            return undefined;
+        }
+
+        // Bound the per-event properties payload. The admin API will reject
+        // oversized events with 422; capping here keeps the in-memory queue
+        // from being filled with multi-MB junk.
+        try {
+            const serializedPropertiesLength = JSON.stringify(event.properties ?? {}).length;
```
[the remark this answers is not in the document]
ydiarra: Not applicable — this was already the case when the review ran. `normalizeEvent` measured with `Buffer.byteLength(JSON.stringify(...), "utf8")`, never `.length`, and the comment above it spells out why: `.length` counts UTF-16 code units and undercounts a CJK payload by ~3x.

Since then the measurement has moved into a shared `serializedPropertiesBytes` helper, so the queue and the control-frame path cannot drift apart on it, and there is a test that feeds it 5000 CJK characters — under the cap by `.length`, three times over it in bytes.

play/src/front/Components/Login/LoginScene.svelte
```diff
@@ -55,6 +55,13 @@
                 "</a>",
         );
     }
+    if (legals?.analyticsNoticeUrl) {
+        legalStrings.push(
+            '<a href="' +
+                encodeURI(legals.analyticsNoticeUrl) +
+                '" target="_blank" class="text-white no-underline hover:underline bold hover:text-white">analytics notice</a>',
+        );
```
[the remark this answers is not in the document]
ydiarra: No longer applicable — the markup this describes is gone. The analytics notice was merged into the world privacy policy on the admin side, so there is no separate URL to link to any more.

Removed: `legals.analyticsNoticeUrl` from `MapDetailsData`, the `legalStrings.push(...)` block from both scenes, and the `analyticsNotice` key from all 15 locales. The sibling comments about `encodeURI` escaping and `rel="noopener noreferrer"` on the same markup are moot for the same reason.

play/src/front/Components/EnableCamera/EnableCameraScene.svelte
```diff
@@ -70,6 +70,13 @@
                 "</a>",
         );
     }
+    if (legals?.analyticsNoticeUrl) {
+        legalStrings.push(
+            '<a href="' +
+                encodeURI(legals.analyticsNoticeUrl) +
+                '" target="_blank" class="text-white no-underline hover:underline bold hover:text-white">analytics notice</a>',
+        );
```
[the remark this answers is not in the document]
ydiarra: No longer applicable — the markup this describes is gone. The analytics notice was merged into the world privacy policy on the admin side, so there is no separate URL to link to any more.

Removed: `legals.analyticsNoticeUrl` from `MapDetailsData`, the `legalStrings.push(...)` block from both scenes, and the `analyticsNotice` key from all 15 locales. The sibling comments about `encodeURI` escaping and `rel="noopener noreferrer"` on the same markup are moot for the same reason.

play/src/front/Connection/ConnectionManager.ts
```diff
@@ -457,6 +463,69 @@ class ConnectionManager {
...
+                    connection.onCleanup(
+                        subscribeToConversationAnalytics(
+                            analyticsConversationStore,
+                            (message) => connection.emitAnalyticsEventReport(message),
+                            {
+                                conversationGroupIdStore: currentPlayerGroupIdStore,
+                                meetingProviderStore: activeMeetingProviderStore,
+                            },
+                        ),
+                    );
```
moufmouf: This code probably does not belong here and is probably too complex.

What I'd do:

- I'd split this in 2 parts: Jitsi on one side and the other communication (a bubble / meeting room in webrtc or Livekit on the other part)
- Jitsi meeting events should be sent in the Jitsi code
- Same for bubble / meeting events
- analyticsClient.meetingStarted and meetingEnded should be promoted to timedEvents

play/src/front/Phaser/Game/MapEditor/Commands/Area/CreateAreaFrontCommand.ts
```diff
@@ -36,6 +37,9 @@ export class CreateAreaFrontCommand extends CreateAreaCommand implements FrontCo
     }
 
     public emitEvent(roomConnection: RoomConnection): void {
+        analyticsClient.mapEditorSaveStarted("area");
         roomConnection.emitMapEditorCreateArea(this.commandId, this.areaConfig);
+        analyticsClient.mapEditorAreaCreated(this.areaConfig.name);
+        analyticsClient.mapEditorSaveSucceeded("area");
     }
```
moufmouf: I have a real doubt about the usefulness of mapEditorSaveStarted and mapEditorSaveSucceeded
Plus the fact that we have 3 analytics calls in the same function is definitely a code smell. Let's maybe keep only one and remove completely mapEditorSaveStarted and mapEditorSaveSucceeded?

play/src/front/Phaser/Game/MapEditor/AreasPropertiesListener.ts
```diff
@@ -1936,9 +1976,21 @@ export class AreasPropertiesListener {
 
             //user in zone to open cowesite with only icon
             inOpenWebsite.set(true);
+
+            analyticsClient.openedWebsite(imageUrl, {
+                targetUrl: this.toCanonicalCowebsiteTargetUrl(initialProperty.link ?? ""),
+                triggerProperty: "openLink",
+                areaId: areaData.id,
+                areaName: areaData.name,
+            });
```
moufmouf: The fact that analyticsClient.openedWebsite is called 4+ times in the code feels weird. Shouldn't we move this down the code to the spot where the website is actually opened and have it only once?

play/src/front/Space/SpacePeerManager/SpacePeerManager.ts
```diff
@@ -184,20 +185,23 @@ export class SpacePeerManager {
...
                 } else if (message.switchMessage.strategy === CommunicationType.LIVEKIT) {
+                    activeCommunicationProviderStore.set("livekit");
                     this._communicationState = new LivekitState(
                         this.space,
                         this._streamableSubjects,
                         blockedUsersStore,
                         this._effectiveScreenSharingLocalStreamStore,
                     );
                 } else {
+                    activeCommunicationProviderStore.set(undefined);
```
moufmouf: Let's put analytics right here instead of going through the activeCommunicationProviderStore that itself triggers the analytics. Less code, more straightforward.

play/src/front/Stores/ScreenSharingStore.ts
```diff
@@ -364,3 +367,31 @@ export const screenSharingLocalVideoBox: Readable<VideoBox | undefined> = derive
...
+        // cancelled in 200ms into a reported second.
+        analyticsClient.screenSharingEnded(screenSharingAnalyticsSessionId);
+    }
+
+    if (!hasStream) {
+        screenSharingAnalyticsSessionId = undefined;
+    }
+
+    previousScreenSharingHadStream = hasStream;
+});
```
moufmouf, line 403: Here, we track local screen sharing. So far, we can only share one item at a time (one screen, one window, etc...) So creating a screenSharingAnalyticsSessionId that is tracked in a map in the AnalyticsClient is useless. screenSharingStarted does not need a screenShareSessionId and in AnalyticsClient, the openScreenShares Map is useless. Simply storing one TimedAnalyticsEventName is enough.

play/src/front/WebRtc/ConversationAnalytics.ts
```diff
@@ -0,0 +1,252 @@
+import type { Readable, Unsubscriber } from "svelte/store";
+import type { AnalyticsEventReportMessage } from "@workadventure/messages";
+import { v4 as uuidv4 } from "uuid";
+import { hasCapability } from "../Connection/Capabilities";
+import type { TimedAnalyticsEventHandle } from "../Administration/TimedAnalyticsEvent";
+import { openTimedAnalyticsEvent } from "../Administration/TimedAnalyticsEvent";
+
+const ANALYTICS_EVENTS_CAPABILITY = "api/analytics/events-batch";
```
moufmouf: This const is squattered 4 times in the code. We could as well make it an exported const and reuse it across files.... or better... drop the code and hard code the string. The hasCapabilities method that takes this string is a Typescript union of all allowed capabilities, so there is no way we can have this string wrong.

play/src/front/WebRtc/ConversationAnalytics.ts
```diff
@@ -0,0 +1,252 @@
...
+    if (conversationType !== "spontaneous_bubble") {
+        return undefined;
+    }
+
+    if (groupId === undefined || groupId === "") {
+        return undefined;
+    }
+
+    return `group:${groupId}`;
+}
```
moufmouf: Let's try to simplify this as much as possible. Ideally, this file is not needed and we just put a meeting.started / meeting.ended in the code at the moment a meeting starts or ends. No need to track changes between Livekit and WebRTC.

play/src/pusher/services/AnalyticsConnectionId.ts
```diff
@@ -0,0 +1,19 @@
...
+ * two ever disagreed on what identifies a socket, the join would mis-attribute
+ * silently — a conversation would land on the wrong session, or none.
+ *
+ * tabId is what keeps two tabs of the same user in the same room apart. When it is
+ * missing the fallback collapses them onto one id, so their intervals interleave
+ * on a single pairing and only one is reported — acceptable, but it is why tabId
+ * is preferred rather than a stylistic choice.
+ */
+export function analyticsConnectionId(socketData: SocketData): string {
+    return socketData.tabId || `${socketData.userUuid}:${socketData.roomId}`;
```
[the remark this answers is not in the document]
moufmouf: Nope. tabId is used to close connections with same tabId that previously connected. So tabId IS unique.
The real question: can it be empty? I don't think so.

play/src/pusher/services/AnalyticsEventsQueue.ts
```diff
@@ -0,0 +1,715 @@
...
+
+    private hasAdminApiConfig(): boolean {
+        return this.endpointUrl !== undefined && this.config.adminApiToken !== undefined;
+    }
+
+    private canSend(): boolean {
+        return this.enabled && this.hasAdminApiConfig() && this.config.maxQueueSize > 0 && this.config.maxBatchSize > 0;
+    }
+
+    private normalizeEvent(
```
moufmouf, line 281: Question: can this now be replaced with a Zod call (if it's not already done?)
I have a feeling the event is already Zoded at this point (through AnalyticsReportMessageHandler.processAnalyticsReportMessage last line that calls `queue.enqueueEvent`

UPGRADE.md
```diff
@@ -1,8 +1,29 @@
-```suggestion
 # Upgrade Guide
 
 This document provides instructions for upgrading WorkAdventure between versions.
 
+## Analytics environment variables renamed (BREAKING CHANGE)
```
moufmouf:
```suggestion
## Upgrading from v1.33.x to v1.34.0

### Analytics environment variables renamed (BREAKING CHANGE)
```

PR #6198, opened 2026-06-08
feat: configurable Room API bind host for dual-stack deployments

### LINKED ISSUE #5871: Room API binds to IPv4 only (0.0.0.0) and can fail when clients use hostname with IPv6
Hi,

while packaging and testing WorkAdventure for self-hosted deployment, I found that the Room API currently binds explicitly to IPv4:

`RoomAPI.bindAsync(`0.0.0.0:${ROOM_API_PORT}`, ...)`

This causes a practical problem on hosts that publish both A and AAAA records.

## Problem

If the public hostname resolves to both IPv4 and IPv6, clients may prefer IPv6 when connecting to the Room API.
In that case, the TCP connection fails because the Room API only listens on IPv4.

In my case:
- hostname had both A and AAAA
- Room API was reachable on IPv4
- Room API connection on IPv6 failed with connection refused
- using the hostname for Room API clients failed unless I forced the IPv4 literal address

## Why this matters

This is not specific to my packaging. It affects any self-hosted setup where:

- the hostname has IPv6 enabled
- clients connect to the Room API via hostname instead of hardcoded IPv4
- address selection prefers IPv6

That makes the Room API effectively IPv4-only by default, even though the deployment itself may be dual-stack.

## What I verified

- DNS returned both IPv4 and IPv6
- Room API initially listened only on 0.0.0.0:<port>
- IPv4 connection worked
- IPv6 connection failed
- after changing the bind host to IPv6, hostname-based connections worked again

## Suggested fix

It would be better not to hardcode 0.0.0.0.

Possible options:

1. Add a configurable bind host, for example:

`ROOM_API_BIND_HOST=[::]`

and then:

`RoomAPI.bindAsync(`${process.env.ROOM_API_BIND_HOST ?? "[::]"}:${ROOM_API_PORT}`, ...)`

2. Or otherwise use a dual-stack-friendly default instead of IPv4-only binding.

## Expected behavior

The Room API should be reachable via the public hostname in dual-stack environments without requiring a hardcoded
IPv4 address.

If helpful, I can also provide the small patch I used in my packaging.

mvanhorn:
## Summary
The Room API server's bind host is now configurable via an environment variable, instead of being hardcoded.

## Why this matters
Requested in #5871: the Room API gRPC server bound to a fixed host, which prevents dual-stack and non-default-interface deployments. A `ROOM_API_BIND_HOST` (validated) environment variable lets operators choose the bind address; the default is unchanged.

## Changes
- Add the bind-host environment variable to the pusher env enum and validator.
- Use it when starting the Room API server, defaulting to the previous value.
- Document it in the self-hosting env-variables reference.

## Testing
The variable is validated through the existing env-variable validator; default behavior is unchanged when it is unset.

Fixes #5871

AI was used for assistance.

### DIFF
--- a/docs/others/self-hosting/env-variables.md
+++ b/docs/others/self-hosting/env-variables.md
@@ -97,7 +97,7 @@ Environment variables for the Play service (frontend and pusher).
 | `SENTRY_RELEASE` | No | Sentry release version identifier for error tracking |
 | `SENTRY_ENVIRONMENT` | No | Sentry environment name (e.g., 'production', 'staging', 'development') |
 | `SENTRY_TRACES_SAMPLE_RATE` | No | The sampling rate for Sentry traces. Only used if SENTRY_DSN is configured. Defaults to 0.1 |
-| `ROOM_API_BIND_HOST` | No | Bind host for the Room API gRPC server. Defaults to 0.0.0.0. Use `[::]` for dual-stack IPv6 deployments. |
+| `ROOM_API_BIND_HOST` | No | Bind host for the Room API gRPC server. Defaults to `[::]`. |
 | `ROOM_API_PORT` | No | Port for the Room API gRPC server. Defaults to 50051 |
 | `ROOM_API_SECRET_KEY` | No | Secret key for Room API authentication |
 | `ENABLE_MAP_EDITOR` | No | Enable the built-in map editor. Defaults to false |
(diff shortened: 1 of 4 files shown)


moufmouf (commented):
Hi @mvanhorn 

Thanks a lot for the PR.
Since `[::]` runs on both IPv4 and IPv6 in most Linux deployments, I think we could safely use `[::]` instead of `0.0.0.0` as the default value. Users wanting to change this can still do it via the environment variable.

I know `[::]` is IPv6 only on Windows, but I don't think anyone is using Windows for WorkAdventure as the default way to deploy it is Linux container.

mvanhorn:
Good call. Done in 483739a: the default is now `[::]`, with the validator description and the env-variables doc updated to match. The bind string that reaches `bindAsync` becomes `[::]:50051`, which is the bracketed IPv6 form grpc-js expects, and `ROOM_API_BIND_HOST` still overrides it for anyone who needs IPv4-only.

moufmouf:
@mvanhorn Thanks! Let's just apply "npm run pretty" to have "Continuous Integration Play" tests passing and we are good to go!

moufmouf:
Just applied the pretty fix. Let's wait for CI and let's merge.

moufmouf:
Merged! Thanks!

mvanhorn:
Appreciate it, @moufmouf - dual-stack deployments can finally bind the Room API where they need it.

PR #6199, opened 2026-06-08
fix: keep long lines in chat code blocks from breaking the layout

### LINKED ISSUE #5688: Very long chat lines break the chat layout
Put a block of code in the chat with a very long line. Check the result.

Image

mvanhorn:
## Summary
Long lines inside chat code blocks now wrap/scroll within the message instead of stretching the chat layout.

## Why this matters
Reported in #5688: a code block containing a long unbroken line widened the whole chat panel, pushing the layout out of shape. Constraining the code block's width keeps long lines contained.

## Changes
- Constrain the code block container in `MessageText.svelte` so long lines stay within the message bounds.
- Minor `Message.svelte` adjustment to support the contained layout.

## Testing
Verified a chat message with a long single-line code block no longer expands the chat panel.

Fixes #5688

AI was used for assistance.

### DIFF
--- a/play/src/front/Chat/Components/Room/Message.svelte
+++ b/play/src/front/Chat/Components/Room/Message.svelte
@@ -134,7 +134,7 @@
 >
     <div
         style={replyDepth === 0 ? "max-width: calc( 100% - 50px );" : "padding-left: 0"}
-        class="flex gap-2 justify-start overflow-visible relative {replyDepth === 0
+        class="flex min-w-0 gap-2 justify-start overflow-visible relative {replyDepth === 0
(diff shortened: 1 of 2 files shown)


moufmouf:
Hi @mvanhorn ,

Thanks a lot for the PR but I think the issue was already fixed (we should have closed #5688 I guess).

When I'm trying with the current version on master, the layout does not break anymore and I can scroll horizontally.

See:

image

Does your PR solves another issue I did not see?

mvanhorn:
Thanks for checking against master. You're right that there's nothing left for this PR to fix, so I'll close it. Glad the layout issue is resolved either way, and #5688 can be closed along with it.

PR #6267, opened 2026-06-24
Improving skipping of tutorial in E2E test

moufmouf:
Skipping tutorial was so far implemented using addLocatorHandler. We are refactoring this to skip the tutorial once the first time we log.


moufmouf:
Ok, useless because we have ENABLE_TUTORIAL=false as an env var

PR #6279, opened 2026-06-26
Add keyboard shortcut to cycle availability status

Arata1202:
## Summary

I often find myself wanting a quick way to change my availability status without opening the profile menu, so this PR proposes a small keyboard shortcut for it.

The new shortcut is `Shift + Y`.

The cycle order is:

`Online -> Busy -> Back in a moment -> Do not disturb -> Online`

The shortcut follows the same lock condition as the profile menu. For example, statuses such as Jitsi, BBB, LiveKit, Silent, and Speaker cannot be overridden by the shortcut.

This also adds the shortcut label to the i18n files.

## Motivation

I would eventually like to contribute more keyboard shortcuts for frequently used actions such as toggling the microphone, camera, or screen sharing.

Before going further in that direction, I wanted to start with a small availability-status shortcut and get feedback on whether this kind of shortcut would be welcome, and whether the chosen key / behavior feels right.

I’m happy to adjust the shortcut key, cycle order, or behavior based on maintainer feedback.

## Checks

- `cd play && npm run typesafe-i18n`
- `cd play && npm run i18n:check`
- `cd play && npm test -- --run tests/front/Rules/StatusRules/requestedStatusCycle.test.ts`
- `cd play && npm run svelte-check`
- `cd play && npm run typecheck`

### DIFF
--- a/play/src/front/Components/ActionBar/AvailabilityStatus/AvailabilityStatusList.svelte
+++ b/play/src/front/Components/ActionBar/AvailabilityStatus/AvailabilityStatusList.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
     import { AvailabilityStatus } from "@workadventure/messages";
     import { resetAllStatusStoreExcept } from "../../../Rules/StatusRules/statusChangerFunctions";
-    import type { RequestedStatus } from "../../../Rules/StatusRules/statusRules";
+    import { requestedStatusChangeLockingStatuses, type RequestedStatus } from "../../../Rules/StatusRules/statusRules";
     import { availabilityStatusStore } from "../../../Stores/MediaStore";
     import { getColorHexOfStatus, getStatusLabel } from "../../../Utils/AvailabilityStatus";
     import LL from "../../../../i18n/i18n-svelte";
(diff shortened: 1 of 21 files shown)


moufmouf:
Hi @Arata1202,

Super idea. A keyboard shortcut for setting the status is a great idea.
Let use just think about it before merging this. I'm wondering in particular if you should not aim for one shortcut per status (instead of cycling).

Pros for cycling:
- you only need to remember one shortcut
Cons for cycling:
- you need to look at the screen while setting the status to end up in the correct status

Taking a look at other solutions, Gather is using one key per status (but they have only 3 status and we have 4) https://support.gather.town/articles/4590630457-keyboard-shortcuts#shortcuts-for-meetings-productivity

@HugoAverty @ydiarra ideas?

Arata1202:
Thanks for the feedback!

That makes sense. I agree that one shortcut per status would make the behavior more predictable than cycling.

Since WorkAdventure has 4 statuses while Gather has 3, I’ll wait for the team’s thoughts on the preferred key mapping. One thing I noticed is that `U` is already used for “Open User List” without modifiers, so it may be worth keeping in mind if we go for a Gather-like layout.

Once the direction is clear, I’m happy to update the PR accordingly.

salah93 (approved):

PR #6292, opened 2026-06-30
feat(meeting): raise hand with ordering

ydiarra:
## What

Adds **raise hand** to meetings, with a server-ordered queue and the moderation controls that go with it.

A participant raises their hand from the action bar; the raised hand shows up

- **above the woka** on the map,
- **on the video tile**, with a **1-based position badge** (order in which hands were raised), and
- in a **raised-hands panel** that appears as soon as there is something to act on.

What the panel offers depends on where you are:

| Context | Who sees the panel | Floor controls |
| --- | --- | --- |
| Megaphone broadcast | admins and genuine zone speakers | **give the floor** (promotes the listener to speaker), **take it back**, **lower a hand** |
| LiveKit meeting room / proximity bubble | everyone | none — everybody already speaks, so the queue is just an ordered list and whoever leads hands over orally |

A participant who was given the floor gets the button back as **give the floor back**. They do *not* inherit the host's moderation rights (a promoted guest cannot promote others).

Each LiveKit meeting area and megaphone listener area can turn the feature off from the map editor (`raiseHandEnabled`, on by default).

## How

The queue is **not** a `SpaceUser` property. It is stored in the **space metadata** under `raisedHands` (and `floorHolders` for who currently holds the floor):

```
click → RaiseHandStore → space.emitUpdateSpaceMetadata({ raisedHands: { raised: true } })
  → pusher (EventProcessor) → back
        Space.applyRaisedHand(senderId, raised)      ← identity = trusted senderId,
        stamps { spaceUserId, name, at }, sorts by at  timestamp + name stamped server-side
  → publishMetadata → every member of the space
        ├─ raisedHandsOrderStore  → tile badge + panel position
        └─ raisedHandPlayerIdsStore → woka indicator on the map
```

Why metadata rather than a flag on `SpaceUser`:

- **The queue is a shared, ordered resource of the space, not an attribute of a user.** Same for the floor-holders list.
- **`SpaceUser` is filtered in broadcast spaces.** `filterOneUser` only dispatches users matching the space's `FilterType`, so in a podium without *See attendees* a speaker never receives the listeners' `SpaceUser` — a flag there would be invisible to the very person who has to act on it. Metadata is broadcast to all members regardless of the filter.
- **One path for every space type.** Proximity bubble, meeting room and megaphone all use the same mechanism, with no interaction with the filter.
- **It stays one message whatever the size of the queue**, instead of fanning out one `SpaceUser` per raised hand.

The client only ever sends its own intent (`{ raised: boolean }` / `{ holds: boolean }`), validated by the shared catalogue; everything that matters — identity, order, timestamp, display name — is stamped server-side. Read-modify-write on the queue is synchronous on the back, so concurrent raises cannot lose an update. A user who disconnects with their hand raised (or while holding the floor) is removed from both lists and the update is broadcast.

Moderation actions are private events (`giveFloor`, `revokeFloor`, `lowerHand`): the queue is owner-writable only, so a moderator clearing the list asks each client to lower its own hand.

## Changes

- **proto**: `GiveFloorPrivateMessage`, `RevokeFloorPrivateMessage`, `LowerHandPrivateMessage`
- **shared-utils**: `RaisedHandsMetadata` (intent + stored schemas), registered in `SpaceMetadataCatalog`
- **back**: `RaisedHandsMetadataProcessor`, `Space.applyRaisedHand` / `applyFloorHolder`, typed `getMetadataValue`, cleanup on user removal
- **pusher**: metadata event processing, private-event processors for the three floor messages
- **front**: `RaiseHandStore`, `RaisedHandsStore`, `RaiseHandAvailabilityStore`, `RaisedHandsAdminVisibleStore`, `RaiseHandZoneSettingsStore`; action-bar button and icon; `RaisedHandsPanel` / `RaisedHandsDock` / `RaisedHandBadge` / `RaisedHandAvatar`; Phaser `UsernameRaisedHandDisplay` + `Character.setRaisedHand`
- **map editor**: `raiseHandEnabled` on `LivekitRoomConfigData` and `ListenerMegaphonePropertyData`, with their editors
- **analytics**: `meeting.hand.toggled`, `meeting.floor.given` / `.revoked` / `.given_back`, `meeting.hand.lowered_for_participant`
- **i18n**: action-bar, map-editor and notification keys across all locales
- **tests**: unit tests on the back processor, the space metadata catalogue and the four front stores; e2e `raise-hand.spec.ts` and `raise-hand-megaphone.spec.ts`

## Notes

- Generated files (`ts-proto`, `typesafe-i18n`) are gitignored and regenerated by CI.


messages/protos/messages.proto
```diff
@@ -1725,6 +1736,10 @@ message SpaceUser{
   google.protobuf.StringValue chatID = 17; // used
   bool showVoiceIndicator = 18;
   bool attendeesState = 19;
+  // Raise-hand state is NOT stored per-user here: it lives in the space metadata key "raisedHands"
+  // (broadcast to all members, incl. a megaphone speaker without seeAttendees). Field numbers 20 and 21
+  // are reserved to avoid reuse.
+  reserved 20, 21;
```
ydiarra: delete

back/src/Model/Space.ts
```diff
@@ -221,6 +221,26 @@ export class Space implements CustomJsonReplacerInterface, ICommunicationSpace {
...
+            // Remove the leaving user from the raised-hands queue (metadata) if present, so it does not keep
+            // a ghost entry, and broadcast the updated queue to everyone.
+            const raisedHandsQueue = this.metadata.get("raisedHands");
+            if (
+                Array.isArray(raisedHandsQueue) &&
+                (raisedHandsQueue as { spaceUserId?: unknown }[]).some((entry) => entry?.spaceUserId === spaceUserId)
+            ) {
+                this.publishMetadata({ raisedHands: this.applyRaisedHand(spaceUserId, false) });
+            }
+
```
ydiarra: use zod to check

back/src/Model/Space.ts
```diff
@@ -221,6 +221,26 @@ export class Space implements CustomJsonReplacerInterface, ICommunicationSpace {
...
+            // Same cleanup for the floor-holders list, so a user who leaves while holding the floor does not stay
+            // in the host's "take back" panel forever.
+            const floorHolders = this.metadata.get("floorHolders");
+            if (
+                Array.isArray(floorHolders) &&
+                (floorHolders as { spaceUserId?: unknown }[]).some((entry) => entry?.spaceUserId === spaceUserId)
+            ) {
+                this.publishMetadata({ floorHolders: this.applyFloorHolder(spaceUserId, false) });
+            }
```
ydiarra: use zod to check

back/src/Model/Space.ts
```diff
@@ -709,6 +729,63 @@ export class Space implements CustomJsonReplacerInterface, ICommunicationSpace {
...
+     * (the back is single-threaded), avoiding lost updates when several listeners raise their hand at once.
+     * The timestamp and name are stamped server-side, and the identity is the trusted senderId.
+     */
+    public applyRaisedHand(senderId: string, raised: boolean): { spaceUserId: string; name: string; at: number }[] {
+        const current = this.metadata.get("raisedHands");
+        const queue = Array.isArray(current)
+            ? (current as { spaceUserId: string; name: string; at: number }[]).filter(
+                  (entry) => entry != undefined && typeof entry.spaceUserId === "string",
+              )
+            : [];
```
ydiarra: use zod to check metadata

back/src/Model/MetadataProcessorInit.ts
```diff
@@ -15,3 +15,17 @@ for (const prefix of proximityQAMetadataPrefixes) {
...
+    const raised = typeof value === "object" && value !== null && (value as { raised?: unknown }).raised === true;
+    return Promise.resolve(space.applyRaisedHand(senderId, raised));
+});
+
+// The client only reports whether it currently holds a granted floor ({ holds: boolean }); the server keeps the
+// authoritative list of floor holders (only users given the floor, never the hosts). See Space.applyFloorHolder.
+metadataProcessor.registerMetadataProcessor("floorHolders", (value, senderId, space) => {
+    const holds = typeof value === "object" && value !== null && (value as { holds?: unknown }).holds === true;
+    return Promise.resolve(space.applyFloorHolder(senderId, holds));
+});
```
ydiarra: use zod to check metadata

moufmouf (commented):
I read the description of the PR, then I read the content of the PR. Those are wildly different. I think I understand why.

The first idea was to put the raised hand flag on the SpaceUser. Super idea because it's super simple.
But... in a podium / audience where people in the audience are not visible, we have an optimization that makes SpaceUsers not dispatched to other users. So if there are 1000 users in a space, and only one speaker, everyone only sees one speaker.

This optimization is great, but it causes a lot of complexity on the code. This PR is a demonstration of that (it could be way easier if the optimization was not here).
I think we should discuss if this optimization is worthwhile or not.

Or we could keep the optimization and make the people who raise their hand dispatched to the other users (with a special state that keeps them hidden from the videoboxcollection?)

PR #6312, opened 2026-07-06
fix: serialize join/leave operations per space in ProximityChatRoomManager

moufmouf:
## Problem

The E2E test **"Should handle rapid transitions between podium and audience zones"** fails intermittently in CI (~10-20% of runs). Analysis of the Playwright traces and server logs from a failing run ([example](https://github.com/workadventure/workadventure/actions/runs/28774384602/job/85315855746)) pinned down a race condition in `ProximityChatRoomManager`:

`joinSpace()` and `leaveSpace()` both mutate the manager's `rooms` map *after* awaiting a server round-trip. When a user crosses quickly from a listener megaphone zone to the matching speaker zone (same space name):

1. The listener zone's `onLeave` starts a full `leaveSpace()`, which awaits the space destruction round-trip (~100-300 ms on loaded CI runners).
2. During that await, the speaker zone's `onEnter` calls `joinSpace()`, which re-uses the room object still present in the `rooms` map. `SpaceRegistry` correctly waits for the in-flight leave, then creates a new space on that room.
3. The leave's continuation then resumes and runs `room.destroy()` + `rooms.delete(spaceName)` — orphaning the room the join just used.

Resulting broken state: the space is alive in `SpaceRegistry` but unreachable from the manager. Every subsequent `joinSpace()` for that space name throws `SpaceAlreadyExistError` (visible in the failing run's console: thrown from `handleSpeakerMegaphonePropertyOnEnter`), and every `leaveSpace()` is a silent no-op (the trace shows the orphaned LiveKit listener participant staying connected until page close). The user can never become a speaker again, so the other participants never see their camera and the test times out — with no possible recovery, which is why both the first attempt and the retry fail identically once the window is hit.

## Fix

Serialize join/leave operations per rooms-map key with a per-space promise chain, mirroring what `SpaceRegistry` already does internally with `leavingSpacesPromises`/`joiningSpacesPromises`. A join for a space can only start once the previous leave for that space has fully completed (including its `rooms` map cleanup), and vice versa. This adds no practical latency: the registry already forced the join to wait for the in-flight leave; now the manager's bookkeeping is ordered consistently with it.

## Tests

Two regression tests reproduce both interleavings with a controlled pending leave/join. Both fail on `master` and pass with this change:

- `does not lose a concurrently re-joined room when a leave completes after its server round-trip` — the exact CI race described above
- `runs a leave requested during an in-flight join only after the join completes` — the mirror-image interleaving

`npm run typecheck`, eslint, prettier and the related front test suites pass.


moufmouf:
Follow-up: replaced the hand-rolled per-space promise chain with map-storage's `LockByKey`, moved to `@workadventure/shared-utils` so both packages share one implementation. Three adjustments were needed: `waitForLock` is now generic (returns the callback's value), the timeout is opt-in (the front usage passes none, because `LockByKey`'s timeout releases the queue while the timed-out callback may still be running — acceptable for map edition, but it would reintroduce this exact race here), and the hard `@sentry/node` import became an injectable `onTimeout` hook since shared-utils is bundled in the browser. map-storage behavior is unchanged (still 10s timeout + Sentry reporting).

moufmouf:
The first CI run on this PR still failed the "rapid transitions" test — but for a new reason, which turned out to be a regression introduced by the serialization itself. Fixed in d83fb3e8b. Details:

**What the CI trace showed:** unlike the original failure (identical `SpaceAlreadyExistError` corruption on attempt and retry), the two attempts now failed at *different* steps, with no error anywhere. In the retry trace, Bob's crossing back from podium to audience produced the leave (LiveKit disconnect) but the re-join never happened: no third "filter added" on the pusher, no LiveKit connect, yet Bob's UI showed the listener state ("Waiting for speaker").

**Root cause:** `GameMapAreas.triggerAreasChange` fires leave-area and enter-area callbacks synchronously in the same call when a single position update carries the player across the one-tile gap between the zones (common at CI frame rates). Before this PR, `manager.leaveSpace()` ran synchronously into `room.leaveSpace()`, which clears `_space` before its first `await` — so a same-batch enter handler saw the room as not joined and did a full join. With the `LockByKey` serialization, the leave's body is deferred to a microtask, so the enter handler's fast-path check (`getCurrentSpaceName() === uniqRoomName` in `AreasPropertiesListener`) now read **stale pre-leave state**, took the "already in this space" branch, and returned without joining. The queued leave then destroyed the space → user stuck with megaphone UI state but no space membership, silently.

**Fix:** the two megaphone enter handlers no longer pre-decide from room state; they always call the serialized `ProximityChatRoomManager.joinSpace()`:
- genuinely still in the space (overlapping zones / role switch): `ProximityChatRoom.joinSpace` early-returns the current space, so no teardown — same cheap role switch as before;
- leave queued (gap crossing): the join runs after it and rebuilds the space — the correct behavior.

The "speaker has priority" check in the listener handler is kept: it reads `activeMegaphoneZones`, which is updated synchronously by the handlers and cannot be stale.

No unit test for this one: the handlers live in Phaser-coupled code with no existing test harness, and the failing E2E test is itself the regression test for this exact sequence.

PR #6320, opened 2026-07-07
fix(play): keep Woka username sharp on Safari when zoomed out

moufmouf:
## Problem

Some users reported that the username label above Wokas is **blurry when zoomed out, specifically on Safari** (fine in Firefox/Chrome).

## Cause

`.username-display` is a promoted compositing layer (`will-change: transform` + `translate3d`). `applyTransform()` then applied `scale(1 / zoomModifier)` to it, which is **> 1 when the map is zoomed out** (`zoomModifier < 1`).

Safari/WebKit rasterizes such a layer once into a cached bitmap and lets the compositor **stretch that bitmap** to satisfy the `scale()`, rather than re-rasterizing the text at the higher effective resolution. Chrome (Blink) and Firefox (Gecko) re-rasterize on scale change, which is why the blur only showed on Safari. The symptom direction matches: blur appears exactly when the transform becomes an up-scale (zoomed out).

## Fix

Remove the `scale()` from the transform (translation only) and fold the zoom factor into the **layout size** instead (`--username-dom-scale`, driving font-size/padding/height/…). The text is now always rasterized at its final resolution and only ever *minified* by the parent Phaser camera transform, which stays sharp on WebKit.

The rendered size is unchanged on all browsers: the leaf previously contributed `zoomModifier·displayScale × (1/zoomModifier) = displayScale`, and now contributes `displayScale × 1` — identical — with the parent camera transform still applying `zoomModifier` as before.

## Testing

- `npm run typecheck` / eslint clean for the changed file (unrelated pre-existing Matrix/tailwind typecheck errors remain in the tree).
- Chrome behavior verified unchanged.
- ⚠️ **Needs verification on real Safari** (deploy env, hence the `deploy` label): confirm the name is sharp both **zoomed out and zoomed in**. Moving the up-scale out of the transform should not reintroduce blur when zoomed in, but this depends on WebKit's backing-store heuristics and should be eyeballed.

### DIFF
--- a/play/src/front/Phaser/Components/UsernameDisplay.ts
+++ b/play/src/front/Phaser/Components/UsernameDisplay.ts
@@ -33,7 +33,13 @@ export class UsernameDisplay {
 
     private readonly onZoomChanged = (zoomModifier: number): void => {
         this.displayScale = this.getDisplayScale(zoomModifier);
-        this.applyStyles();
+        // The ancestor scale we cancel in applyStyles()/applyTransform() is measured from the DOM
+        // layer, so invalidate it and re-apply at render time, once Phaser has updated the layer's
+        // transform for the new zoom. Applying immediately would measure the previous frame's scale.
+        this.gameScene.usernameDomLayer.invalidateAncestorScale();
+        this.scene.events.once(Phaser.Scenes.Events.RENDER, () => {
+            this.applyStyles();
+        });
     };
     private toForeFront: boolean = false;
     private depth: number = 0;
@@ -65,6 +71,14 @@ export class UsernameDisplay {
         this.updateUsernameBackgroundColor(outlineColor);
 
         this.gameScene.usernameDomLayer.addUsername(this.element);
+
+        // The immediate applyStyles() above may have measured the ancestor scale before Phaser
+        // applied the layer transform for this frame. Re-apply once at render time so the initial
+        // scale is measured against the real on-screen transform.
+        this.gameScene.usernameDomLayer.invalidateAncestorScale();
+        this.scene.events.once(Phaser.Scenes.Events.RENDER, () => {
+            this.applyStyles();
+        });
 
         this.scene.game.events.on(WaScaleManagerEvent.ZoomChanged, this.onZoomChanged);
     }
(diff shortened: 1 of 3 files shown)


moufmouf:
### Update — the first commit was blurry when zoomed *in*

Testing showed the initial fix (translation-only transform, all scaling in layout) fixed the zoomed-out Safari blur but made the name **blurry when zoomed in on every browser** (Chrome/Firefox/Safari).

Reason: a browser derives a promoted layer's rasterization resolution from the element's **own** transform, not from an ancestor's. The first commit handed all the scaling to the parent Phaser camera transform, so when zoomed in every browser upscaled the leaf's cached bitmap.

**Follow-up commit** keeps the scale on the element's own transform but **clamps it at 1** (`Math.min(1/zoomModifier, 1)`) so it only ever *minifies*; the leftover enlargement when zoomed out is baked into the layout size. Net result:
- Zoomed **in**: identical to the original known-good code (own transform minifies → sharp everywhere).
- Zoomed **out**: transform scale clamped to 1, enlargement in layout → sharp on Safari.

Rendered size is unchanged at every zoom level. Please re-check on the deploy env: name should be sharp **both zoomed in and zoomed out**, on Safari and Chromium/Firefox.

moufmouf:
### Update 2 — medium-zoom blur on low-DPI screens (dual-monitor clue)

Testing showed the clamped version was still blurry on Safari at a **medium zoom level, and only on a 2K/non-Retina monitor** — crisp on the Retina panel of the same laptop at the same zoom.

**Root cause:** a Safari promoted layer is crisp only when its total scale (own transform × ancestors) is ≤ 1. The ancestor here is the Phaser DOM layer, scaled by **`actualZoom`**, and `actualZoom = realSize.width / gameSize.width / devicePixelRatio` — it divides by DPR. My previous commit cancelled `zoomModifier` in the element's own transform, but the real ancestor factor is `actualZoom`. They agree at the extremes (so zoomed in/out looked fine) but diverge at medium zoom, where the own transform is clamped to exactly 1 and has no headroom: `actualZoom` then goes **> 1 on a 1× display** (parent upscales the cached bitmap → blur) but stays **≤ 1 on a 2× Retina display** (→ crisp).

**Fix:** cancel `actualZoom` instead of `zoomModifier` in the own transform (`Math.min(1 / getActualZoom(), 1)`). Size-neutral — rendered size is `displayScale × actualZoom` either way. Also broadened `emitZoomChangedIfNeeded` to fire on `actualZoom` changes, so the name refreshes on resize / moving the window between screens of different DPI (not just on `zoomModifier` changes).

Please re-test on the deploy env, specifically **sweeping through medium zoom on the 2K monitor**, and dragging the window between the two monitors.

moufmouf:
### Update 3 — root cause found via on-device measurements

The `actualZoom` change was a no-op at the broken spot. The measurements you captured settled it:

| case | dpr | ownScale | ancestorScale | totalScale | result |
|---|---|---|---|---|---|
| mid, 2K | 1 | **1.0** | 1.687 | 1.687 | blurry |
| very zoomed-in, 2K | 1 | 0.247 | 8.093 | 2.002 | sharp |
| mid, Retina | ~2 | 1.0 | 1.022 | 1.022 | sharp |

Two facts overturned my model:
1. **"total scale > 1 → blur" is false** — the zoomed-in name is sharp at total scale 2.0 while the blurry one is only 1.69. The real discriminator is the **own transform scale**: Safari rasterizes a promoted layer at full resolution only when its own transform is a *minification* (< 1). When it's a pure translation (`scale === 1`) Safari caches at `devicePixelRatio` and lets the parent stretch the bitmap → blurry at 1×, fine at 2×. My clamp `min(1/x, 1)` produced `ownScale === 1` across the whole zoomed-out/mid range — I built the dead zone.
2. The parent's real on-screen scale (`ancestorScale = 1.687` at mid) is **neither `zoomModifier` nor `actualZoom`** (both ≤ 1 there), so it can't be derived — hence the no-op.

**Fix (commit `1c95c13d4`):** measure `ancestorScale` from the DOM via a hidden fixed-size probe in the username container (`rect.width / layoutWidth`), and cancel it in the own transform: `min(1 / ancestorScale, 1)`. This makes the own transform a genuine minification whenever the parent magnifies (total scale ≤ 1), so Safari rasterizes at full resolution. Size is unchanged. Measurement is lazy/cached, invalidated on zoom+resize, and read at render time.

Please re-test the **mid zoom on the 2K monitor** (and a slow zoom sweep + dragging between monitors). If you can, re-run the console snippet at the previously-blurry mid zoom — I'd expect `ownTransformScale` ≈ 0.59 now (not 1.0) and `totalScale` ≈ 1.0.

PR #6331, opened 2026-07-08
fix(livekit): display the sound meter (volume indicator) in video bubbles

ydiarra:
## Problem

In a proximity bubble, each video box shows a 7-bar **sound meter** that reflects how loudly the person is speaking. It works on the peer-to-peer (simple-peer) path but stays **frozen** for **LiveKit** participants.

## Root cause

The meter is driven by `Streamable.volumeStore`. In `LivekitParticipant`, both streamable builders hardcoded it:

```ts
// getVideoStream() and getScreenShareStream()
volumeStore: writable(undefined),
```

So `volumeStore` is permanently `undefined` → `SoundMeterWidget` renders static bars. The simple-peer path (`RemotePeer`) instead builds a `derived` store that runs a `SoundMeter` (Web Audio `AnalyserNode`) over the remote audio stream every 100 ms.

The audio `MediaStream` needed to feed a `SoundMeter` is **already available** on the LiveKit streamable (`_audioStreamStore` for the camera, `_audioScreenShareStreamStore` for screen share) — it was just never wired to an analyser.

## Fix

- Export the existing `createVolumeStore()` helper from `MediaStore.ts` (guards on an audio track, spins up a `SoundMeter`, polls `getVolume()` every 100 ms, cleans up).
- In `LivekitParticipant`, build two `derived` volume stores over the LiveKit audio streams (camera + screen share) and use them instead of `writable(undefined)`, mirroring the WebRTC path.

The `derived` store is lazy: the `SoundMeter` is created only when a `VideoMediaBox` subscribes (widget mounted), and is rebuilt/stopped when the audio stream changes (mute/unmute) or the widget unmounts. Since LiveKit already plays remote audio through its own `<audio>` elements, the audio track is live and the `AnalyserNode` receives real data.

No change to the WebRTC path, the local meter, or the UI components — they already consume `volumeStore` identically.

## Test plan

- [ ] Two clients in a **LiveKit**-backed world join the same bubble → the 7-bar meter animates on the remote participant's video box (as it does on WebRTC).
- [ ] Mute / unmute the mic → the meter stops, then resumes.
- [ ] Screen share **with audio** → the meter animates too.
- [ ] Non-regression: WebRTC (simple-peer) bubbles and the local meter behave as before.

### DIFF
--- a/play/src/front/Components/Video/VideoMediaBox.svelte
+++ b/play/src/front/Components/Video/VideoMediaBox.svelte
@@ -3,7 +3,7 @@
     import { getContext, onDestroy, onMount } from "svelte";
     import * as Sentry from "@sentry/svelte";
     import type { Subscription } from "rxjs";
-    import SoundMeterWidget from "../SoundMeterWidget.svelte";
+    import SoundMeterWidgetWrapper from "../SoundMeterWidgetWrapper.svelte";
     import { highlightedEmbedScreen } from "../../Stores/HighlightedEmbedScreenStore";
     import type { VideoBox } from "../../Space/VideoBox";
     import { LL } from "../../../i18n/i18n-svelte";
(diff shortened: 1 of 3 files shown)


moufmouf (commented):
The volume computation is typically an operation that can take some CPU. I remember an article about Jitsi scaling stating that this was actually the bottleneck to scale to 500 users.
In our case, we should ensure the volume is computed only if the videomediabox is visible. I'm not sure if this is the case or not already. We need to check this.

PR #6343, opened 2026-07-10
perf(livekit): batch video subscription changes to cut renegotiation churn

moufmouf:
> Stacked on top of #6342 (auto-mute). Base branch is `feat/auto-mute-microphone-crowded-meeting`, so this PR's diff shows only Part B.

## What & why

The post-#6342 trace showed the remaining main-thread cost is a **~4s renegotiation burst on meeting join** (then it settles to ~0). Root cause: every `RemoteTrackPublication.setSubscribed()` immediately sends an `UpdateSubscription` signal, which the SFU answers with a `mediaSectionsRequirement` that forces a publisher-side renegotiation. livekit-client only debounces its publisher `negotiate()` by ~20ms, so video subscription changes that dribble out across animation frames — many tiles mounting on join, or a Woka moving past a crowd — each cause their own renegotiation.

## How

New `VideoSubscriptionBatcher`: camera/screen-share `setSubscribed()` calls are routed through a shared batcher that applies all pending changes in one synchronous flush (default **50ms** window):

- **Cluster the renegotiations** — flushing together makes the SFU's media-section updates land inside livekit's 20ms negotiate debounce, so a burst collapses into a **single** renegotiation instead of one per frame.
- **Cancel transient churn** — a tile that becomes visible then hidden within the window resolves to "no change" and never hits the network.
- **Batch across participants** — one shared instance, so a join storm flushes as one batch.

Audio is untouched (subscribed once, never toggled). The existing 75ms unsubscribe grace is preserved; only the final `setSubscribed()` call is now batched. Trade-off: video appears up to 50ms later — imperceptible.

## Tests

- New `VideoSubscriptionBatcher.test.ts` — 7 cases: batched (not synchronous) apply, transient-churn coalescing, net-state of a burst, redundant-request skip, multi-publication single pass, fresh window after flush, and error isolation (one throwing publication doesn't block the batch).
- The 4 existing `LivekitParticipant` subscription tests updated for the batched timing.
- 16 tests pass; typecheck / lint / prettier / svelte-check all clean.

## How to validate

Capture a fresh Chrome trace on join into a populated meeting: the per-second renegotiation burst (createOffer/setLocalDescription) during the first ~4s should shrink further vs the #6342 trace.


moufmouf:
Note: it does not seem to work at all.....

PR #6376, opened 2026-07-20
test(e2e): stabilise pre-existing flaky E2E (Firefox WebGL, media readiness, map editor)

ydiarra:
## What & why

The E2E suite intermittently **red-fails on pre-existing flakiness** that is unrelated to whatever PR surfaces it. Investigating three affected PRs (#6363, #6356, #6370) showed **none of them are regressions** — the failures all trace back to shared test infrastructure. This PR fixes those root causes so the suite stops flaking for every PR.

### 1. Firefox can't create a WebGL context in CI → the dominant blocker
From the Playwright traces, **6/6 Firefox failure traces** show:

```
WebGL creation failed … Exhausted GL driver options (WEBGL_EXHAUSTED_DRIVERS)
Phaser v4.2.0 (Canvas | Web Audio)
```

Firefox headless on the runner fails to get a WebGL context, so **Phaser 4 falls back to the Canvas renderer**, which can't bring the game scene up. The game never becomes ready, so the `microphone-button` load-gate (`expect(...).toBeVisible({ timeout: 120_000 })`) times out. This is Firefox-only (chromium has software WebGL via ANGLE), and it also underlies the **map-editor flakes** (`map_editor_open_website`, `map_editor_upload_entities`) which run over the degraded canvas.

**Fix:**
- `tests/playwright.config.ts` (firefox + mobilefirefox): `webgl.force-enabled`, `webgl.disable-fail-if-major-performance-caveat`, `gfx.webrender.software` so Firefox accepts a software WebGL context.
- CI workflow: a firefox-only step installs `libgl1-mesa-dri` (llvmpipe) and exports `LIBGL_ALWAYS_SOFTWARE`/`GALLIUM_DRIVER=llvmpipe`.

### 2. Media-button readiness race in `createUser`
`createUser` asserts the mic/camera buttons reach `normal` state with the default 10s expect timeout. Acquiring the fake devices can take longer on Firefox CI; a miss also prevents the auth storage state from being cached, so the whole shard re-runs `createUser` and re-rolls the flake. **Fix:** 30s timeout on those checks (`Menu.expectButtonState` gains an optional `timeout`).

### 3. Load diagnostics
`getPage` now logs `pageerror`/`crash` so a stalled load surfaces the underlying JS error instead of an opaque 120s timeout.

### 4. `picture_in_picture` "not available" (webkit)
The PiP button only renders while `$isInRemoteConversation`, which lags peer visibility on webkit. **Fix:** 10s → 20s, matching the connection waits.

### 5. `AreaEditor.addProperty` clicked with no readiness wait
A bare `getByTestId(property).click()` timed out (20s action timeout) on the 2nd drawn area. **Fix:** wait visible + `scrollIntoViewIfNeeded` before clicking.

## Notes
- Behaviour-preserving: no fixed sleeps, no masking — a genuinely stuck button / missing element still fails.
- The WebGL change only affects Firefox rendering in CI and **can only be validated by re-running the E2E jobs** — please re-run the firefox/mobilefirefox jobs a couple of times to confirm the flake rate drops.

### DIFF
--- a/.github/workflows/build-test-and-deploy.yml
+++ b/.github/workflows/build-test-and-deploy.yml
@@ -223,13 +223,18 @@ jobs:
           pactl info
       - name: Enable software WebGL for Firefox
         if: ${{ contains(matrix.project, 'firefox') }}
-        # Headless Firefox on the runner otherwise exhausts its GL driver options, WebGL
-        # creation fails, and Phaser 4 falls back to the Canvas renderer — which cannot bring
-        # the game scene up, so the game never becomes ready (the 120s load-gate flake that
-        # intermittently red-fails E2E). Provide the Mesa software rasteriser and force its use;
-        # the firefox projects also set webgl.force-enabled so Firefox accepts the software context.
+        # Headless Firefox on the runner exhausts its GL driver options, WebGL creation fails,
+        # and Phaser 4 falls back to the Canvas renderer — which cannot bring the game scene up,
+        # so the game never becomes ready (the 120s load-gate flake that intermittently red-fails
+        # E2E). Providing Mesa's software rasteriser + prefs helps but is not 100% reliable in
+        # pure headless mode, so we run Firefox HEADED under a virtual X display (Xvfb): it then
+        # gets a stable GLX + llvmpipe WebGL context. WA_FIREFOX_HEADED gates the headed mode to
+        # exactly these jobs (via playwright.config.ts) so no other job is affected.
         run: |
-          sudo apt-get install --yes libgl1-mesa-dri
+          sudo apt-get install --yes libgl1-mesa-dri xvfb
+          Xvfb :99 -screen 0 1920x1080x24 -ac +extension GLX +render -noreset > /tmp/xvfb.log 2>&1 &
+          echo "DISPLAY=:99" >> "$GITHUB_ENV"
+          echo "WA_FIREFOX_HEADED=1" >> "$GITHUB_ENV"
           echo "LIBGL_ALWAYS_SOFTWARE=true" >> "$GITHUB_ENV"
           echo "GALLIUM_DRIVER=llvmpipe" >> "$GITHUB_ENV"
       - name: 'Setup .env file'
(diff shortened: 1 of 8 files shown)


ydiarra:
## ✅ Validated — full green E2E run

Root cause of most of the recurring E2E flakiness: **Firefox fails to create a WebGL context in headless CI**. It exhausts its GL driver options, Phaser 4 falls back to the Canvas renderer, the game scene never becomes ready, and the `microphone-button` load-gate times out (the `120_000ms` flake). Confirmed from the Playwright traces — every Firefox failure trace showed `WebGL creation failed … Exhausted GL driver options` followed by `Phaser v4.2.0 (Canvas)`. This also explains the map-editor flakes (`open_website`, `upload_entities`, `megaphone_speaker_zone`) that run over the degraded canvas.

What this PR does:
- Runs Firefox **headed under a virtual X display (Xvfb)** in CI (+ Mesa `llvmpipe` + `webgl.force-enabled` prefs) so it gets a stable WebGL context. Gated by `WA_FIREFOX_HEADED`, set only in the firefox E2E jobs that start Xvfb, so no other job is affected.
- Hardens shared flaky spots without hiding real failures (no fixed sleeps): the login-path media-button check (`expectButtonState` gains an optional timeout), `AreaEditor.addProperty` (wait visible + scroll before click), the webkit picture-in-picture wait (10s→20s), the `scripting_audio_stream` audio wait (20s→40s), and `variables "storage works @docker"` (`expect.poll` instead of a single read after container reboots).
- Adds `pageerror`/`crash` diagnostics to `getPage` and `createUser`.

**Latest run: all 14 E2E jobs green** — chromium ×4, firefox ×4, webkit ×3, mobile ×3 (run 29833695716). Failure traces from earlier runs now show `Phaser (WebGL)` instead of Canvas.

Known residual: an **intermittent** mobilefirefox page-crash at login (did not reproduce in the green run) — now instrumented via the `createUser` crash listener so the next occurrence surfaces the underlying error.

PR #6381, opened 2026-07-21
refactor(ui): migrate the hard-case design-system usages onto components

ydiarra:
> Re-opens #6372, which was merged into `refactor/design-system-components` by mistake instead of `master`. That branch has since landed on master (#6378), so this PR is rebased onto `master` and contains **only** the hard-case changes (the redundant components-migration commit was dropped in the rebase).

Follow-up to #6378 (ex-#6371), which deliberately left a set of design-system usages raw because they needed more than a mechanical swap. This PR clears **almost all** of them — the two points I flagged as "reversible, ask me".

## Point 1 — usages that needed the component adapted

- **`.avatar` (2)**: recreated `UI/Avatar.svelte` with an `id` prop and migrated the two boxes (`Message`, `TypingUsers`). Imported **under an alias** so it sits beside the domain `Chat/Components/Avatar.svelte` without a name clash.
- **`.input-search` (1)**: recreated `Input/InputSearch.svelte` as a **bare `<input>`** (no wrapper div) so `MapList` keeps its own `relative` container and the `peer`-driven search icon keeps reacting to focus/hover. Added a `type` prop to preserve `type="text"`.

## Point 2 — buttons whose scoped CSS was the blocker → `:global()`

The reason these couldn't be `<Button>` before: Svelte scopes CSS by hashing elements in the component's own markup, and a component's internal `<button>` doesn't get that hash. Hoisting each selector to `:global()` fixes it.

- 6 `responsive-message` PopUps, plus `MatrixChatSettingsModal`, `TutorialInstruction`, `ErrorScreen`, `Jitsi`/`LivekitRoomPropertyEditor`. Where a rule referenced a `@keyframes`, the keyframe is marked `-global-` so the global rule can still reach it.
- `SelectCompanionScene` **keeps** its `button {}` rule (8 other raw buttons still need it); the 2 migrated buttons get a `pointer-events-auto` utility instead.
- `GlobalCommunicationModal`'s `button.light[disabled]` became `disabled:*` utilities, dropping the generic `.light` hook rather than globalizing such a broad class name.
- `VisitCard`'s (dead — the button is a flex item) `.visitCard button { float }` kept faithfully via `:global()`.

## Also

- **`Button` gains a bindable `element`** — `EmojiSubMenu` uses it as its floating-UI anchor, clearing the last `bind:this` blocker.
- **`PwaInstallScreen`** drops `variant="light"`: its translucent button needs white label text, which the DS's `.btn-light .btn-label` (dark) would have broken once the text moved into the `.btn-label` span. No-variant keeps the base white label.

## Left raw, reported

- **`WarningBanner`** — its admin-configurable **inline** text colour must reach the text node, which `.btn-label` would override. No clean path without a bespoke Button API; not worth it for one banner.
- The only other raw `btn` occurrences left in the codebase are **inside HTML comments**.

## Verification

`svelte-check` (`--fail-on-warnings`), `eslint`, `prettier` all clean on the changes. **All 285 test ids preserved** (before/after diff). No behaviour or wording changes.

**Not verified visually** (build/vitest can't run on the dev host). This is the change most worth eyeballing on the `deploy` env — several buttons had scoped CSS whose `:global()` behaviour I reasoned about but couldn't render.

### DIFF
--- a/play/src/front/Chat/Components/MatrixChatSettingsModal.svelte
+++ b/play/src/front/Chat/Components/MatrixChatSettingsModal.svelte
@@ -3,6 +3,7 @@
     import { scale } from "svelte/transition";
     import type { MatrixChatConnectionLike, MatrixUserSettingsDiagnostics } from "../Connection/ChatConnection";
     import ButtonClose from "../../Components/Input/ButtonClose.svelte";
+    import Button from "../../Components/UI/Button.svelte";
     import LL from "../../../i18n/i18n-svelte";
     import { clearIgnoredSuggestedRooms } from "../Stores/ChatStore";
     import { DEBUG_MODE } from "../../Enum/EnvironmentVariable";
@@ -226,12 +227,13 @@
         {#if !loading && data}
             <div class="flex flex-shrink-0 flex-col gap-2 border-t border-white/10 p-4 sm:p-5">
                 <p class="text-xs text-white/55">{$LL.chat.matrixSettings.clearSuggestedRoomsHint()}</p>
-                <button
+                <Button
                     type="button"
-                    class="btn transition-all duration-200 active:scale-[0.98] disabled:opacity-50 disabled:active:scale-100 {successFlash ===
-                    'clearSuggested'
-                        ? 'btn-success matrix-settings-btn-success'
-                        : 'btn-light btn-border'}"
+                    variant={successFlash === "clearSuggested" ? "success" : "light"}
+                    appearance={successFlash === "clearSuggested" ? "filled" : "border"}
+                    class={`transition-all duration-200 active:scale-[0.98] disabled:opacity-50 disabled:active:scale-100 ${
+                        successFlash === "clearSuggested" ? "matrix-settings-btn-success" : ""
+                    }`}
                     disabled={syncing}
                     onclick={handleClearSuggested}
                 >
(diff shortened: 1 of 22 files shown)


ydiarra:
The failing E2E jobs here are pre-existing shared-infra flakiness, **not** a regression from this PR:

- **firefox 3/4** — `map_editor_megaphone_speaker_zone` (property-panel click), `map_editor_open_website` (Google-apps URL), `map_editor_upload_entities` (PDF drop), and a 120s `microphone-button` load-gate. These are the map-editor / WebGL-load flakes.
- **mobilefirefox 1/1** — `matrixChat` `Cancel edit message` / `Retrieve encrypted message` / `Key creation` failing on the shared login checks (`bg-danger` mic/camera state + the 120s mic-button gate).

Why it's not this PR (even though it touches `Components/UI/Button.svelte`):
- The mic/camera buttons the `bg-danger` assertions target (`MicrophoneMenuItem`/`CameraMenuItem`) aren't modified here.
- The `Button.svelte` diff doesn't change the plain `bg-danger` class.
- All chromium / webkit / desktop-firefox jobs pass the same `bg-danger` assertions — a real class regression would fail chromium too.

The dominant root cause is Firefox not obtaining a WebGL context in headless CI (Canvas fallback → the game never loads). All of these are addressed by #6376, which now produces a fully green E2E run. **Rebasing this branch onto master once #6376 is merged should clear these failures.**

PR #6443, opened 2026-08-19
perf(ci): shard the end-to-end matrix by measured duration

ydiarra:
The workflow takes about 30 minutes on a green run and the end-to-end matrix is almost all of it. This brings the critical path to roughly 15 minutes on a PR, for about the same runner minutes.

Measured on run [31467905058](https://github.com/workadventure/workadventure/actions/runs/31467905058) (green, 30:08), by parsing the `list` reporter output of all 14 legs.

## What was slow

**Sharding balanced the wrong thing.** Playwright's `--shard` splits by number of tests. The four chromium shards held 63/55/56/48 tests but took 22.5/13.7/11.0/6.3 minutes, so the matrix waited on a shard doing three times the work of its neighbour.

**Shard counts did not follow project weight.** webkit had three shards for 3.4 minutes of tests — legs of 1.3 minutes paying 3.2 minutes of setup — while mobilechromium and mobilefirefox had none for 13 and 14 minutes.

**The setup was fully sequential** even though its two halves are independent: the npm and Playwright installs (~55 s) do not depend on the `docker compose` boot (116 s, mostly image pulls).

## What changed

`--shard` is replaced by `--test-list`, filled by `tests/scripts/shard-list.mjs`, which packs whole spec files by their recorded duration (longest-processing-time first). Durations live in `tests/shard-timings.json` and are regenerated with `npm run e2e:timings`.

| project | shards | longest leg (tests only) |
|---|---|---|
| chromium | 4 → **6** | 22:30 → **8:49** |
| firefox | 4 → **5** | 13:20 → **8:21** |
| webkit | 3 → **1** | 3:25 |
| mobilechromium | 1 → **2** | 13:00 → 6:18 |
| mobilefirefox | 1 → **2** | 14:20 → 6:55 |
| mobilewebkit | 1 | 0:50 |

On the reference data every shard of a project now lands within three seconds of the others.

`tests/tests/chat/matrixChat.spec.ts` (21 tests, 7.0 min) was split by theme in the first commit. Packing is per file, so the heaviest file is a floor on the shard duration, and this one was above the target. The floor is now `map_editor_megaphone_speaker_zone.spec.ts` at 4.3 min.

Per-leg setup drops from ~3:12 to ~2:12: the stack starts in the background right after the `.env` is written and a later step collects it and propagates its exit code, using the `everything_started` gate that already existed in `docker-compose.e2e.yml`. Adds the npm and Playwright browser caches, which these jobs never had, updates `checkout@v2`/`setup-node@v3` to v4, and installs one browser per mobile leg instead of all three (`browser: ""` was installing everything).

## The trap this had to guard

Playwright resolves `--test-list` entries against `testDir` and **ignores the ones it cannot match**: a list gone stale runs zero tests and still exits 0. A silent green. The job therefore resolves the list through Playwright before running it (`shard-list.mjs --verify`) and fails when it selects nothing.

## Verified

Locally: the partition covers all 77 spec files with no duplicate for each of the six projects at the chosen shard counts; `--verify` accepts a real shard and rejects a stale list; the error paths (shard out of range, more shards than files, unknown project, >20% unrecorded files) behave; eslint and prettier clean on `tests/`; `tsc` and 13 unit tests green on the `e2e-report` workspace.

Not verified: the workflow itself. The backgrounded compose, the caches, and the real durations are only proven by this PR's own run. Worth checking on it:

- the tests executed across the 17 legs still add up to ~607 — the real proof that no file was dropped
- the chromium and firefox legs land between 8 and 9 minutes
- `Wait for WorkAdventure` takes clearly less than the 116 s + 55 s it used to cost in sequence

## Notes

- This is a **PR** win. On master, `run-gitlab-saas-tests` (17–27 min, external GitLab pipeline) becomes the new critical path at ~28 min.
- `prod-single-domain-deploy-tests` and `helm-test` still use `--shard`. They are label-gated and off the PR critical path, so they were left alone.
- Out of scope by decision, and still on the table: `workers: 1` in `playwright.config.ts` leaves three of four cores idle; ~136 s of hardcoded sleeps per project pass (50 s of it in two `setTimeout` calls in `livekit.spec.ts`); no path filters, so a docs-only PR runs all 17 legs.
- Heads-up: the split moves the two SAS emoji verification tests into `matrixChatVerification.spec.ts`, which will conflict with any in-flight work on that flow.


moufmouf:
Superseded by #6543 which does not rely on test-lists. If de-synchronized, no tests are skipped.

PR #6472, opened 2026-09-02
feat(front): reroute or stop a pathfinding move when the collision grid changes mid-walk

ydiarra:
Stacked on #6466 (base branch `fix/locked-area-pathfinding`; GitHub will retarget to `master` once it merges).

## The gap left by #6466

#6466 made the pathfinding grid honor blocked areas, but the grid is snapshotted when `GameScene.moveTo()` starts and the path is never reconsidered. Lock a meeting room *while someone is already walking towards you* (chat's **Talk to**, a right click, `WA.player.moveTo`): they enter anyway. Path following runs with `setDirectControl(true)` and re-applies a `setPosition()` per frame computed from the precomputed path, so the area's collider is overridden frame after frame.

## Behavior

- On each tile change during path following (each waypoint shift in `Character.followPath` — EasyStar emits one waypoint per tile), the **next waypoint** is checked against the **live** collision grid (`getCollisionGrid` is cached and lazy — no rebuild unless something actually changed). Only the next waypoint, on purpose: the woka keeps walking up to the obstacle, so the blocked warning always shows next to the area it talks about, and a lock lifted before the woka reaches the area never interrupts the walk.
- The next waypoint started colliding → the path is aborted right before the blocking tile and `GameScene.moveTo()` computes a new route to the **original destination**. If one exists, the woka follows it **silently**.
- Destination not reachable anymore → the woka stops where it is (right at the border) and the blocked-area warning is displayed above it via `Area.displayBlockedWarningMessage()` — the exact same message and admin unlock affordance as a physical collision with the area. No new i18n keys.
- A cancellation from the user taking over (keyboard input, a new right click, teleport) resolves **without** the internal `blocked` flag and never triggers a reroute.
- `MAX_PATH_REROUTES = 5` caps pathological lock/unlock flapping.
- The scripting contract of `WA.player.moveTo` is unchanged: `{ x, y, cancelled }`.

## Implementation notes

- `Character.followPath` gets a protected `onPathWaypointReached()` hook; only `Player` overrides it. `RemotePlayer` is untouched — its movement mirrors server state and must never be cut.
- `PathFollowResult` gains an optional `blocked` flag, consumed inside `GameScene.moveTo()` and stripped from what callers see. `finishFollowingPath` gains an optional `blocked` parameter, forwarded by the two overrides.
- Stored waypoints are collider-adjusted (y shifted up by the body offset); the tile mapping inverts that offset exactly (`play/src/front/Utils/PathValidation.ts`, pure and Phaser-free).
- `Area.displayWarningMessageOnCollide` is now the public `Area.displayBlockedWarningMessage`, shared between physical collisions and the pathfinding stop.
- A player already **inside** an area when it locks never gets a false abort: their grid does not mark the area (`getAreaBlockReason` returns `locked` only when the player is outside), so their exit path stays valid.

## Tests

- `play/tests/front/Utils/PathValidation.test.ts`: pure vitest cases (green, 4/4).
- One E2E test in `map_editor_lock_area.spec.ts` chaining both mid-walk scenarios on a lockable wall, driven by a slow `WA.player.moveTo(x, y, speed)` (the `speed` param passes through unclamped, making mid-walk locking deterministic): stop-at-the-border + warning (destination at the wall's vertical middle so even the nearest-available fallback finds no reachable tile), then unlock and silent reroute around the wall.

## Verified locally

- Typecheck: 0 errors in the touched files (remaining worktree noise is the pre-existing stale `@workadventure/messages` build, unrelated).
- ESLint: clean on all touched files (play linted in a Linux container).
- Vitest: 4/4.
- E2E: CI runs the full suite on this PR.

## Still out of scope

Enforcement remains 100% client-side: the back returns any user's position on request and never validates a destination. A modified client can still go anywhere.

### DIFF
--- a/play/src/front/Phaser/Game/GameScene.ts
+++ b/play/src/front/Phaser/Game/GameScene.ts
@@ -194,7 +194,6 @@ import type { ChatConnectionInterface, ChatUser } from "../../Chat/Connection/Ch
 import { selectedRoomStore } from "../../Chat/Stores/SelectRoomStore";
 import { raceTimeout } from "../../Utils/PromiseUtils";
 import { PLAYTEXT_NEW_MEDIA_DEVICE_PREFIX } from "../Entity/Character";
-import type { PathFollowResult } from "../Entity/Character";
 import { type Avatar, ConversationBubble } from "../Entity/ConversationBubble";
 import { DarkenOutsideAreaEffect } from "../Components/DarkenOutsideArea/DarkenOutsideAreaEffect";
 import { isInsidePersonalAreaStore } from "../../Stores/PersonalDeskStore";
@@ -3886,11 +3885,10 @@ ${escapedMessage}
     ): Promise<{ x: number; y: number; cancelled: boolean }> {
         const pathfindingManager = this.getPathfindingManager();
         const resolvedSpeed = speed ?? this.CurrentPlayer.walkingSpeed;
-        let lastBlockedResult: PathFollowResult | undefined;
 
         // The collision grid can change while the path is followed (an area gets locked or full...): the
-        // player then aborts the path (see Player.onPathWaypointReached) and we compute a new route to the
-        // destination.
+        // player then aborts right before walking onto a blocking tile (see Player.onPathWaypointReached)
+        // and we compute a new route to the destination.
         for (let attempt = 0; attempt <= MAX_PATH_REROUTES; attempt += 1) {
             pathfindingManager.setCollisionGrid(
                 this.gameMapFrontWrapper.getCollisionGrid({ emitMapChangedEvent: false }),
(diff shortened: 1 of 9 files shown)


play/src/front/Phaser/Game/GameScene.ts
```diff
@@ -3881,18 +3885,112 @@ ${escapedMessage}
...
+        const messageId = `path-blocked-${blockingArea.id}`;
+        this.CurrentPlayer.destroyText(messageId);
+        this.CurrentPlayer.playText(
+            messageId,
+            message,
+            5000,
+            () => this.CurrentPlayer.destroyText(messageId),
+            true,
+            "warning",
         );
```
moufmouf: Shall we try to reuse the code of `Area.displayWarningMessageOnCollide` here (or mutualize this code somehow in a shared method?)
Otherwise the day there is a new block reason, we might modify it in one place and forget the other place.
Plus the one in `Area.displayWarningMessageOnCollide` is already a better implementation because it allows you to unlock the locked room if you are an admin.

play/src/front/Utils/PathValidation.ts
```diff
@@ -0,0 +1,23 @@
...
+    colliderTileType: number,
+): number {
+    return waypoints.findIndex(
+        (waypoint, i) =>
+            i > 0 &&
+            grid[Math.floor((waypoint.y + waypointYOffset) / tileDimensions.height)]?.[
+                Math.floor(waypoint.x / tileDimensions.width)
+            ] === colliderTileType,
+    );
+}
```
moufmouf: I think we don't need to search on all future waypoints if one of them is locked.
We just need to check the NEXT waypoint.

Indeed, if I'm very far from the room and suddenly, someone locks the room, I'll get a message displayed saying the room is locked. But I'm not even close to the room, so it's hard for the user to understand what room is locked.

If we check only the next waypoint, the user will keep walking to the locked room until it reaches it. So when the message displays, it's obvious which room is locked. Furthermore, if the room is locked and unlocked while the use is walking to the room, the user is not stopped and will reach the room.

PR #6477, opened 2026-09-03
feat: Add Vietnamese (vi-VN) translation

ydiarra:
## Description

This PR adds a complete Vietnamese (vi-VN) locale to the play front, following the same pattern as the Traditional Chinese PR (#6216): the locale folder plus the locale-detection test fixtures (the generated `i18n-util*` files are no longer tracked).

- All 30 translation files in `play/src/i18n/vi-VN/`, complete against `en-US`
- `vi-VN` added to the mocked locale lists in `play/tests/front/Utils/locales.test.ts`
- `index.ts` deep-merges over `en-US` so any future missing key falls back to English

Translation notes:
- Neutral polite address ("bạn") throughout
- `randomNames` template reordered to `{name} {adjective}` to match Vietnamese word order
- Keyboard hints localized as `[PHÍM CÁCH]` / `[ESC]`
- Untranslated on purpose: product names (WOKA, Jitsi, Matrix…), image/video asset paths, the Lorem ipsum placeholders present in `en-US`

The translation is LLM-assisted; a review by a native Vietnamese speaker is very welcome before merge.

## Checks

- `npm run typesafe-i18n` picks up the new locale
- `npm run i18n:check` → ✅ All translations are complete
- Prettier clean

PR #6478, opened 2026-09-03
feat: Add Thai (th-TH) translation

ydiarra:
## What

Adds a complete **Thai (th-TH)** locale to the play frontend, following the same structure as the Korean (ko-KR) locale (#5367):

- `play/src/i18n/th-TH/` — all 29 namespace files + `index.ts` (deepmerge over en-US), 100% key coverage.
- A `"th"` browser-detection case in `play/tests/front/Utils/locales.test.ts`.

No fallback font is bundled: Thai glyphs render with the system Thai fonts (Thonburi / Leelawadee / Noto), which is fine since there is no global `font-family` rule. The Noto Sans Thai fallback that was in earlier revisions of this PR was dropped as unnecessary.

No registration code needed anywhere: locale folders are auto-discovered by `typesafe-i18n`, the language selector uses `Intl.DisplayNames`, and generic `th` is auto-mapped to `th-TH` by the navigator detector. Thai is LTR, so the `ar-SA` RTL special-case is untouched.

## Translation notes

- Product nouns kept as-is (Woka, WorkAdventure, Jitsi, Matrix…); megaphone → เมกะโฟน; bubble → บับเบิล.
- `randomNames` culturally adapted with Thai noun+adjective order (`template: "{name}{adjective}"`).
- `recording.expireIn` keeps the `{s}` param inline after the unit word, same as the ja/ko/zh locales (the generated `Translation` type requires all base params).
- ⚠️ Machine-assisted first draft — **a native Thai review is very welcome** before/after merge.

## Verification

- `npm run typesafe-i18n` regenerates cleanly with th-TH in the union.
- `npm run i18n:check` passes (exit 0, all locales complete); reverse diff confirms no extra keys.
- Targeted `tsc --noEmit` on the th-TH folder passes.
- `vitest tests/front/Utils/locales.test.ts`: 10/10 passing, including the new `th` case.
- Rebased on master after the Vietnamese (vi-VN) locale landed (#6477); both locales are listed in the test.


play/src/front/Phaser/Components/Ui/Button.ts
```diff
@@ -37,7 +37,7 @@ export class Button extends Container {
         this.text = this.scene.add
             .text(0, 0, "", {
                 color: "0x000000",
-                fontFamily: '"Press Start 2P"',
```
moufmouf, line 40: Do we still use Press Start 2D?
Maybe we should consider refactoring this part of the code to use our modern design (if this part of the code is not dead)

play/src/front/style/font.css
```diff
@@ -3,6 +3,9 @@
 @import "@fontsource/roboto-condensed";
 @import "@fontsource/press-start-2p";
 @import "@fontsource-variable/oswald";
+/* Fallback font for Thai characters (subsets ship with unicode-range: U+0E00-0E7F) */
+@import "@fontsource/noto-sans-thai";
+@import "@fontsource/noto-sans-thai/500.css";
```
moufmouf: I think we need to double check the weight of this font. It won't be needed for most users so we need to ensure it's not loaded if it's not used (based on the unicode-range)

play/src/i18n/th-TH/onboarding.ts
```diff
@@ -0,0 +1,56 @@
...
+        description: "แชร์หน้าจอของคุณกับคนอื่น ๆ ในบับเบิลสนทนา เหมาะสำหรับการนำเสนอและการทำงานร่วมกัน!",
+        video: "./static/images/screensharing.mp4",
+        hint: "คลิกปุ่มแชร์หน้าจอที่ไฮไลต์ไว้เพื่อเริ่มแชร์!",
+        next: "ถัดไป",
+    },
+    pictureInPicture: {
+        title: "ภาพซ้อนภาพ (Picture in Picture)",
+        description:
+            "ใช้โหมดภาพซ้อนภาพเพื่อให้วิดีโอคอลยังมองเห็นได้ขณะที่คุณเดินสำรวจแผนที่ เหมาะสำหรับการทำหลายอย่างพร้อมกัน!",
+        video: "./static/Videos/PictureInPicture.mp4",
```
moufmouf, line 44: Untranslated strings should be removed completely (so every file name in those translations that is not a Thai screenshot should be removed)

PR #6495, opened 2026-09-04
feat(front): let a window that starts in the background load its world

ydiarra:
## The problem

A window that starts hidden — launched at login, minimised while a world loads, a background
browser tab — never reaches its world. Chromium schedules no `requestAnimationFrame` for a hidden
renderer, Phaser's loop is rAF-driven (`forceSetTimeOut: false`), and `connectToRoomSocket` has a
single caller, reached from `GameScene.create()`. No frame → no scene switch → no `connect()` → no
room. The loading screen sits there until the user comes back, and only then does the world begin
to load.

## What this PR does

When frames stop arriving, `App.svelte` takes Phaser's loop over and clocks it by hand from a
`setInterval` until the world is reached. **Same code path as always — only the clock changes — so
no ordering invariant is touched.** One file, ~45 lines.

## Design notes

- **One driver at a time.** While the interval drives, Phaser's own clock is put to sleep and woken
  the moment a frame arrives. `sleep`/`wake`/`tick` are `TimeStep`'s documented methods and `wake`
  is idempotent, so nothing internal is touched. Considered and rejected: replacing
  `game.loop.raf` with a custom scheduler — an undocumented seam plus a fallback path.
- **Trigger is frame staleness, not `document.hidden`.** In Electron a child `WebContentsView`
  reports `document.hidden === true` while rendering at 120fps, until a first hide/show cycle.
  Anything gated on the visibility API would pump non-stop from launch.
- **Bounded on both ends.** Nothing is pumped while frames flow; it stops for good once the world
  is reached, or at a 60s ceiling for a user parked on a name/woka screen that no tick can get
  past. Stopping always hands the clock back, or a window revealed later would never redraw.
- **Why a page timer and not a worker.** The first commit clocked the loop from a blob-URL worker,
  on the theory that Chromium's background throttling would starve a page timer. Testing the plain
  `setInterval` showed it gets the boot to the room from a hidden window, so the second commit
  drops the worker, its module and its test. Note that Chrome driven through CDP (Playwright) does
  not throttle background tabs at all, so this cannot be measured from the e2e suite.
- **Why not `backgroundThrottling: false`?** Tried on the desktop app and reverted: one
  un-throttled `webContents` un-throttles the whole window, forever, for every open world. And it
  forces the page to believe it is always visible, which `visibilityStore` — driving video
  mounting and PiP — would act on.

## Not covered

- A user with no name or woka still waits on the interactive screen; no clock gets past it.
- The deeper fix — a room session that does not belong to a rendering scene — is a separate
  project; an earlier attempt broke three ordering-dependent consumers and is preserved on
  `feat/join-before-render`.

Supersedes the background-loading part of #6458, which also carries a network prefetch and a
`GameMapFrontWrapper` refactor; those can follow on their own.


play/src/front/Components/App.svelte
```diff
@@ -209,6 +209,53 @@
...
+        let driving = false;
+        const handBack = () => {
+            if (driving) {
+                driving = false;
+                loop.wake();
+            }
+        };
+        const watchFrames = () => {
+            lastFrameAt = performance.now();
+            handBack();
```
moufmouf: Shouldn't we call clearInterval(bootPump) here?

If a frame is displayed, we should handBack AND we can stop the setInterval altogether as it's become useless.

moufmouf (commented):
Question: shouldn't we move all the bootPump code in a separate file called from App.svelte to keep App.svelte clean and small?

play/src/front/Components/App.svelte
ydiarra: Deliberately not, and I've written the reason into the code so it doesn't come up again.

A frame only proves the window is visible *at that instant*. On a window that boots visible, the first rAF fires at ~t+16ms — clearing the interval there re-opens the second case this PR is about: a world minimised *while it loads* loses its clock and freezes again.

So the frame callback keeps doing `handBack()` (Phaser takes its own clock back), but the interval stays armed. While frames flow it ticks nothing — one comparison every 16ms — and it stops for good on the only two conditions that make it useless: the world is reached, or the 60s ceiling.

The pump now lives in `play/src/front/Phaser/Services/BackgroundBootPump.ts` (your other comment), with that note above `watchFrames`.

PR #6496, opened 2026-09-04
test(e2e): wait for the lock broadcast before pathfinding moves in the lockable area tests

ydiarra:
## Problem

`map_editor_lock_area.spec.ts` › "Lock area also blocks pathfinding moves of a user allowed to edit the map" is flaky on firefox: it failed twice in a row on #6488's CI (run 33886112972) and once (then passed on retry) on run 33884686268 of an unrelated branch.

The Playwright trace shows the race: Alice's lock button loses `bg-danger` when the server broadcast comes back to her page, and Admin1's `moveTo` starts ~30 ms later, before the same broadcast has been applied on his page. His collision grid still sees the area as locked. Since #6480 such a move walks up to the locked area and resolves without error instead of rejecting, so the assertion reads `x = 0`.

The same window exists in the other direction (lock, then move) and in the mid-walk test after the unlock.

## Fix

- `e2eHooks.getCollidingAreasCount()`: number of areas colliding with the local player (locked, full, no access).
- `Map.waitForCollidingAreasCount(page, count)`: polls that hook.
- The lockable area tests wait for the moving player's page to reflect each remote lock toggle before starting the pathfinding move.

No product code path changes.

### DIFF
--- a/play/src/front/Utils/E2EHooks.ts
+++ b/play/src/front/Utils/E2EHooks.ts
@@ -311,6 +311,14 @@ export function decrementLivekitRoomCount() {
 }
 
 /**
+ * Number of areas the local player currently collides with (locked, full, no access...). Lets a test wait until a
+ * lock toggled from another browser has reached this client before starting a pathfinding move.
+ */
+function getCollidingAreasCount(): number {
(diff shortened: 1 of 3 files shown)


moufmouf (changes requested):
That's one extra method in `map.ts` and one extra method in `E2EHooks.ts` just for this test. I'd rather add a small timeout than so much code for just one E2E test.

PR #6501, opened 2026-09-07
feat(recording): tell the admin when an egress ends

ydiarra:
## What

The back notifies the admin API when a LiveKit recording egress **ends**, so the admin can turn a completed one into a customer-facing webhook (`recording.completed`, see the SaaS webhooks feature).

- `POST /api/recordings/events` on the admin, only when `ADMIN_API_URL` is set. Payload: `phase` (always `ended`), `status` (`EGRESS_COMPLETE`, `EGRESS_FAILED`, …), `egressId`, `recordingSessionId`, `playUri`, `recorder` (`uuid`, `spaceUserId`), ISO `startedAt` / `endedAt`, `error`, and the files LiveKit wrote (`filename`, `sizeBytes`, `durationSeconds`). The admin reads the status to tell a usable recording from a failed one.
- Retried 3 times on transport / 5xx errors (250 ms, 1 s, 4 s); a 4xx is not retried. Fire-and-forget from `CommunicationManager`: a failing admin never delays or breaks the recording flow (logged + Sentry).
- `HandleRecordingWebhookRequest` gains `startedAtMs`, `endedAtMs` and `fileResults`, copied from the LiveKit `EgressInfo` (nanosecond bigints divided before conversion). Used only inside the back.
- Nothing is sent when an egress starts.

## Why

Today the admin only ever discovers recordings by listing the S3 bucket once a day. Customers asked to be notified on their own server when a recording is ready, with a download link.

## Tests

- `back/tests/CommunicationManager.test.ts`: notifier called once with the full payload when an egress ends, not called on start nor for an unknown session, a rejecting notifier does not throw.
- `back/tests/LivekitService.test.ts`: timestamps and file results copied (and zero/empty when absent).
- `back/tests/AdminApi.test.ts` (new): URL + token, retry then give up, no retry on 4xx.
- `npm run typecheck`, `npm run lint` and `npm run pretty-check` clean.

### DIFF
--- a/back/src/Services/AdminApi.ts
+++ b/back/src/Services/AdminApi.ts
@@ -5,7 +5,63 @@ import * as Sentry from "@sentry/node";
 import { ADMIN_API_TOKEN, ADMIN_API_URL } from "../Enum/EnvironmentVariable";
 import { LivekitCredentialsResponse } from "./Repository/LivekitCredentialsResponse";
 
+/**
+ * What the admin learns when a recording egress starts or ends. It turns
+ * this into customer webhooks (recording.started / completed / failed).
+ */
+export interface RecordingEventPayload {
+    phase: "started" | "ended";
+    status: string;
+    egressId: string;
+    recordingSessionId: string;
+    playUri: string;
+    recorder: { uuid: string; spaceUserId: string };
+    startedAt: string | null;
+    endedAt: string | null;
+    error: string | null;
+    files: { filename: string; sizeBytes: number; durationSeconds: number }[];
+}
+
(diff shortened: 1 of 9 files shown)


back/src/Services/AdminApi.ts
moufmouf, line 1: We need to document this new endpointi in Swagger.

Take a look at the AdminApi.ts in pusher, we have "@openapi" annotations.

Note: it is likely we will have to put the "@openapi" annotations in the pusher and not in the back because Swagger lives on the pusher.

moufmouf (changes requested):

PR #6534, opened 2026-09-11
Leave a codec for a cheaper one when the encoder is CPU-limited

moufmouf:
Stacked on #6523 (Phase 0). Phase 1a of the adaptive codec design: the client reacts to a sustained CPU limitation by moving one codec rung down. Phase 1b (the back moving a P2P bubble to LiveKit when a CPU-limited member encodes for several peers) comes in a follow-up PR on top of this one.

## What it does

`qualityLimitationReason === "cpu"` in the outbound stats means libwebrtc's overuse detector has **already** lowered the resolution or the frame rate. The machine is never unprotected; the question is whether that degradation is the right one for the viewers. For a software encoder, a cheaper codec at full resolution usually is, above all on a screen share where sharpness is the product, at the price of bandwidth.

`CpuLimitationDetector.ts` samples the aggregated encoder stats of each category (the numbers the local feedback tile shows) once a second:

- 60-sample window, 70 % threshold: Jitsi's 60 s streak, with gaps tolerated and lone spikes ignored. The first 10 samples of a stream are skipped (keyframes and rate-control ramp-up look like overload).
- The window restarts when the stream stops, the transport changes or the codec changes. One decision per window, then a window of cooldown.
- Never for H.264 (nothing below), never for a hardware encoder (a cheaper codec gains it nothing). The camera holds back while a screen share is being sent: the screen share is the heavy encoder and goes first.

A decision demotes the codec for the rest of the session (`demotedCodecStore`): `preferredVideoCodecs()` leaves it out with every codec above it, so screen share goes AV1 → VP9 → H.264 and camera VP9 → H.264. Nothing persisted, nothing goes back up before a reload: the browser's own smooth history (Phase 0 probe) is the memory across sessions.

The publishers subscribe to the store and reconfigure themselves through the code Phase 0 already routes through:

- **LiveKit**: the codec is fixed at `publishTrack()`, so the room unpublishes and publishes again, the two updates a share that stops and starts goes through. The camera publication is normally reused for the whole room (turning it off pauses it, because of a memory leak seen in unpublish/publish cycles, see 55a76a3b0): a demotion is the one time per session it is really unpublished.
- **P2P**: `RemotePeer` sets the fork's `receiveCodecs` to the new negotiable list and calls `negotiate()`; the bitrate budget follows the new codec on `negotiated`. Transceivers and ICE stay: a keyframe and a sub-second freeze for the viewers.

`@workadventure/simple-peer` 12.3.0 re-applies the codec preference before every offer, as it already did before answers, so a changed preference holds whoever initiates the renegotiation (fork test added for both cases).

Each demotion is sent to PostHog as `wa_codec_downgrade`. Firefox reports no limitation reason and never demotes.

## Tests

- `CpuLimitationDetector.test.ts`: warm-up, threshold, gaps and spikes, cooldown, resets, one sample a second whatever the number of encoders, H.264/VP8/hardware bypass, camera held back.
- `VideoPresets.test.ts`: a demotion removes the codec and those above it, per category.
- `LiveKitRoom.test.ts`: screen share and camera republished with the next codec; later camera toggles keep the pause path.
- Fork `test/codec-preferences.ts`: a changed preference applies to both directions when the initiator or the non-initiator renegotiates.

To see it by hand: DevTools → Performance → CPU 6× slowdown, screen share on quality "high" (1440p AV1), video quality stats overlay on: `Limited by: cpu`, then about 70 s later the codec changes. Worth checking the heap before and after a **camera** demotion, to size the unpublish leak.

Docs: `docs/others/internals/bandwidth-management.md`, "When the encoder cannot keep up".


play/src/front/WebRtc/CpuLimitationDetector.ts
```diff
@@ -0,0 +1,130 @@
...
+     * `screenShareRunning` holds the camera back: the screen share is the heavy encoder, it goes first, and demoting
+     * the camera in the meantime would mark the wrong codec.
+     */
+    public sample(
+        stats: LocalEncoderStats | undefined,
+        screenShareRunning: boolean,
+        now: number = Date.now(),
+    ): VideoCodec | undefined {
+        // "P2P (3 encoders)" and "P2P" are the same transport
+        const transport = stats?.source.split(" ")[0];
```
moufmouf: This analysis based on a string here is dangerous.

Maybe LocalEncoderStats should contain the raw data (so "sources" instead of "source", or stats.encoder[0].source... and maybe we could instead rely on the Svelte component to actually merge this data into one? Or even better, the Svelte component could display raw data for some encoders, like:

"Codec: VP9/AV1/VP9"

moufmouf: Done in 7c2d37e2f and 51a3b125a.

You were right, and the string turned out to be the least of it. `mimeType` and `encoderImplementation` in the aggregate both come from the *most limited* encoder, and which encoder that is changes from second to second. On a P2P bubble whose peers negotiated different codecs, the summary codec flipped as the worst peer changed and reset the 60-sample window every time — the detector could never fire; when it did, it left a codec belonging to one peer of three. The hardware bypass was the same coin flip, vetoing or not depending on who was worst at that instant.

So `LocalEncoderStats` now carries `encoders`, the readings it summarises (replacing `encoderCount`), and the detector decides on the set: it demotes only a codec every encoder is on, and skips only a set that is hardware all the way through (a partly-hardware set still has software encoders to relieve). `source` is no longer decorated with "(3 encoders)", so the `split(" ")` is gone, and the tile appends the count from `encoders.length`.

And the tile now shows the codecs the way you suggested — `Codec: VP9/AV1/VP9`, one per encoder, `video/` prefix dropped since it is noise once there are three of them. That display is also what tells you why the detector is holding back on a mixed set.

PR #6535, opened 2026-09-11
Move a bubble to LiveKit when a member's encoders cannot keep up with several P2P peers

moufmouf:
Stacked on #6534 (Phase 1a), itself on #6523 (Phase 0). Phase 1b of the adaptive codec design: the transport lever, pulled before the codec one.

## Why

In P2P, Chrome runs one video encoder per `RTCPeerConnection`: with three peers a weak machine encodes every camera frame three times. On LiveKit it encodes once (SVC or simulcast add a fraction, not a multiple), and the viewers lose nothing. Dropping AV1 → VP9 halves the encode cost; going from three encoders to one divides it by three. So when the CPU limitation detector of #6534 fires on encoders serving more than one P2P peer, the first action is now to ask for LiveKit, and the codec is only demoted if that does not help.

## What it does

- **Client**: the detector's first action in P2P with several peers raises `cpuLimitedStore`, which `SpacePeerManager` synchronises to the space as a new `cpuLimited` field of `SpaceUser`, like the camera state. The flag is remembered for the session and never lowered: lowering it once on LiveKit, where the load is gone, would bounce the bubble back and forth. The transport switch resets the detector's window; if the back cannot switch, the next window falls through to the codec demotion. Reported to PostHog as `wa_cpu_limited_flag`.
- **Back**: `TransitionPolicy.shouldTransition()` takes the number of flagged users. WebRTC → LiveKit when a flagged user is present **and** the bubble has more than two users (a pair gives one encoder each, LiveKit saves nothing there, and this is what makes the session-long flag safe for the next 1:1). LiveKit → WebRTC is vetoed by any flagged user **whatever the count**, the way a recording pins the bubble. The asymmetry is deliberate: once a bubble moved for a flag, only the flagged user leaving brings it back, so a third member coming and going never bounces it. That leave takes the existing 20 s delayed path, and a return within 20 s cancels it, as today.
- A `SpaceUser` field update never evaluated the policy (`handleUserUpdated` was a pass-through): it now does when the update mask carries `cpuLimited`, leaving every other update as before.
- **Self-hosters**: `LIVEKIT_SWITCH_ON_CPU_LIMITATION` (back, default `true`) turns it off for those who would rather not pay SFU bandwidth for bubbles that would have stayed peer-to-peer.

## Tests

- `TransitionPolicy.test.ts`: flag below the threshold with 2 and 3 users, LiveKit unavailable, no fallback while flagged (even for a pair), fallback once the flag is gone, env var off.
- `CommunicationManager.test.ts`: a `cpuLimited` update evaluates the policy with the flagged count and triggers the immediate transition; other updates still do not.
- `CpuLimitationDetector.test.ts`: flag before codec with several P2P peers, once per session, hardware encoders included; straight to the codec for a single peer or on LiveKit.

- `tests/webrtc_cpu_limitation.spec.ts` (opt-in, `RUN_CPU_LIMITATION_E2E=1`): one page reports every outbound video stream as CPU-limited, so the whole chain runs for real on a deployment. Passed against this branch's environment: the bubble of three moved to LiveKit on Alice's flag, stayed there after Carol left, and Alice's camera was republished in H.264 on LiveKit (3.3 min). With `CPU_LIMITATION_EXPECT=demote` it passed against #6534's environment: Alice and both peers renegotiated H.264 in P2P.
- A fix in `tests/utils/auth.ts`: the cached login state is now reused only for the deployment it was saved against (storage is per origin), which is what made the first runs against a second environment sit on the name prompt.

By hand: three users in a P2P bubble on a deployment with LiveKit, CPU 6× slowdown on one of them with the camera on, video quality stats overlay on: after about 70 s the bubble moves to LiveKit (the tile's source changes from P2P to Livekit) and stays there when the third user leaves, until the throttled user leaves.

Docs: `docs/others/internals/bandwidth-management.md`, "When the encoder cannot keep up"; the env var is documented in `.env.template`, `docker-compose.livekit.yaml` and the generated env variables page.


back/src/Model/CommunicationManager.ts
```diff
@@ -326,14 +334,26 @@ export class CommunicationManager implements ICommunicationManager {
...
+        if (!this.policy.shouldTransition(currentType, userCount, cpuLimitedUserCount)) {
             this.orchestrator.cancelPendingTransition();
         }
     }
 
+    /**
+     * What the policy decides on: the current transport, how many users are in the space, how many of them raised
+     * their cpuLimited flag.
+     */
+    private policyInputs(): [CommunicationType, number, number] {
```
moufmouf: Actually, the call is always:

```
const [currentType, userCount, cpuLimitedUserCount] = this.policyInputs();
// Then
this.policy.shouldTransition(currentType, userCount, cpuLimitedUserCount)
```

We could have a local private CommunicationManager.shouldTransition() method with no parameters than inlines this.policyInputs and then calls this.policy.shouldTransition

PR #6544, opened 2026-09-13
feat(play): add DEFAULT_CAMERA_PRIVACY_SETTINGS and DEFAULT_MICROPHONE_PRIVACY_SETTINGS

### LINKED ISSUE #6541: Feature request: Server-side default for camera/microphone 'away mode' privacy settings
### Is your feature request related to a problem?

When a user leaves the browser tab (goes "away"), WorkAdventure keeps the microphone enabled by default (`microphonePrivacySettings` defaults to `true` in `play/src/front/Connection/LocalUserStore.ts`). In organizations that care about privacy, this means a user's microphone can stay hot while the tab is in the background and there is currently **no way for a self-hosted instance administrator to change this default for all users**.

### Describe the solution you'd like

A server-side configuration (env variable on the Play service, e.g. `DEFAULT_CAMERA_PRIVACY_SETTINGS` / `DEFAULT_MICROPHONE_PRIVACY_SETTINGS`, or a map/world-level setting) that overrides the client-side defaults in `LocalUserStore`:

- `cameraPrivacySettings` currently defaults to `false` (camera disabled in away mode). This is good.
- `microphonePrivacySettings` currently defaults to `true` (microphone stays enabled in away mode). This is the one we would like to be able to force to `false` for all users by default.

The relevant code (`LocalUserStore.ts`, ~line 572):

```ts
getMicrophonePrivacySettings() {
    if (localStorage.getItem(microphonePrivacySettings) == null) {
        localStorage.setItem(microphonePrivacySettings, "true");
    }
    return localStorage.getItem(microphonePrivacySettings) === "true";
}
```

Ideally, when such an env variable is set, the user's personal setting would either be forced (admin wins) or simply used as the *initial* default until the user changes it themselves.

### Describe alternatives you've considered

- Patching `LocalUserStore.ts` and building a custom Play image. Works, but must be maintained on every upgrade and only affects first login (existing users keep their stored value).
- Injecting JS via reverse proxy to overwrite localStorage. It breaks the ability of users to opt in themselves.
- Asking users to manually disable the option in the settings menu. This option is not scalable.

### Additional context

- WorkAdventure is used in a school/public-sector context where audio privacy compliance is required.
- A per-user setting to opt back in (if the "admin wins" variant is chosen) would still be valuable.

0xr2po:
Closes #6541

## Summary

DEFAULT_CAMERA_PRIVACY_SETTINGS & DEFAULT_MICROPHONE_PRIVACY_SETTINGS env variables

Allow self-hosted instance administrators to configure the initial value of the camera and microphone privacy settings ('keep device enabled when the tab is away') for users who have not set their own preference yet.

The values are exposed on the Play service as environment variables, sent to the client through the map details payload (like skipCameraPage) and used as in-memory defaults in LocalUserStore. Users can still change the settings in the settings menu; their personal choice always wins over the admin default.

Defaults are unchanged (camera: false, microphone: true). Means no change from current deployment without touching env.


moufmouf (changes requested):
Hi @0xr2po ,

Thanks a lot for the PR. It's pretty good idea!

We'll need 2 changes before merging this.

**First change:**

Right now, you are storing the default values (`defaultMicrophonePrivacySettings` and `defaultCameraPrivacySettings`) in the `LocalUserStore`. This class is really dedicated to storing data in localStorage. So default values (that are not stored in local storage) should not be stored here.

Instead, I'd rather have you follow the typical way we store data coming from the MapDetailsData, i.e. in the `Room` class.

**Second change:**

You change is working for people doing self-hosting (without an Admin) but will stop working for people that use the AdminAPI. We need to keep the symmetry here. The `LocalAdmin` class (used when there is no Admin API) has a sister class named `AdminApi` that implements the same `AdminInterface`. In `AdminApi.fetchMapDetails`, it would be great to be able to fetch those same settings (`defaultMicrophonePrivacySettings` and `defaultCameraPrivacySettings`) from the AdminApi.

Thanks again for this PR!

play/src/pusher/services/AdminApi.ts
```diff
@@ -294,7 +296,14 @@ class AdminApi implements AdminInterface {
...
-                return mapDetailData.data;
+                // Fall back to the environment variables to keep symmetry with LocalAdmin.
+                return {
+                    ...mapDetailData.data,
+                    defaultCameraPrivacySettings:
+                        mapDetailData.data.defaultCameraPrivacySettings ?? DEFAULT_CAMERA_PRIVACY_SETTINGS,
+                    defaultMicrophonePrivacySettings:
+                        mapDetailData.data.defaultMicrophonePrivacySettings ?? DEFAULT_MICROPHONE_PRIVACY_SETTINGS,
+                };
             }
```
moufmouf, line 307: No need to fallback to the default value here. Let's just use what the AdminAPI sends (or the front will fallback to the old behaviour).
DEFAULT_CAMERA_PRIVACY_SETTINGS and DEFAULT_MICROPHONE_PRIVACY_SETTINGS will be ignored if there is not AdminAPI and this will match the behaviour of the other properties.
```suggestion
                return mapDetailData.data;
```

moufmouf:
LGTM! Let's merge this, thanks a lot!

PR #6550, opened 2026-09-14
feat(analytics): say which meeting a microphone or speaking period happened in

ydiarra:
Stacked on #6549 — review that one first; this PR's diff is the single commit on top. Its base is `fix/megaphone-ghost-meeting`, deliberately: without that fix the megaphone space is a "meeting", and every speaking period in a world with a megaphone configured would be attributed to it.

## What is missing today

`media.microphone.dwell` and `media.speech.dwell` already carry an interval each, measured by the pusher — one row per speaking turn (with a 1500 ms hold so ordinary pauses do not cut a sentence into fragments), with both ends and a duration.

What they do not carry is **which meeting it happened in**. Their `openProperties` is `z.object({})`. So the rows that would let a meeting be replayed — who had their microphone open, who spoke when — or two participants be related to each other, can only be reached by joining on overlapping time ranges across everyone who was in the room at the time.

## What this changes

`meetingId` on both events, optional: a microphone can be open, and someone can talk, outside any meeting at all.

**Volume is unchanged.** These rows are already emitted, one per interval; this adds a field to them. The timeline was never a question of how much to record — only of whether the recorded rows could be related to anything.

The id comes from a small module (`CurrentMeeting.ts`) that the two owners of a `meeting.ended` interval now write to — `SpacePeerManager` for bubbles and meeting areas, `JitsiMeetingAnalytics` for Jitsi — so there is exactly one answer to "which meeting is this tab in", and it is the same string the meeting row carries.

Two details worth a look in review:

- **Clearing is guarded on the id.** A bubble dissolving just after a meeting area was entered would otherwise erase the meeting that is actually live. Covered by a test.
- **A period that straddles a meeting boundary is cut at the boundary**, not attributed whole. Microphones stay open across several meetings, so without that the field would be decided by wherever the user happened to be when they unmuted. Speaking periods get this for free — the factory reads the current meeting per interval.

No `roomId` beside it: the envelope already carries `roomId` as a column of its own.

## What it unlocks, without further client work

- **The timeline of a meeting** — `WHERE event_name IN ('media.speech.dwell', 'media.microphone.dwell') AND meetingId = ?`, ordered by `startedAt`.
- **The relation graph** — who was in a meeting with whom, and for how long, from a self-join on `meetingId`; and who spoke around whom, from overlapping speaking intervals.

Both are queries, not new storage. On the admin side they need a materialized `meeting_id` column and a skip index before they are viable against the 2 s ClickHouse query timeout, since `meetingId` lives in `properties_json` and is not in the sort key — that is a separate MR on the SaaS repository.

## A caveat to state plainly

Speaking time is a loudness estimate from the speaker's own analyser, not recognition — no audio is inspected, transmitted or stored. It is also the most sensitive thing this pipeline produces: "who spoke, for how long, with whom" is nominative data about employees. It rides the existing `collaboration_activity` consent category today, which is probably not the right home for a relation graph; that question belongs with whoever exposes it, not with this PR.

## Tests

`play/tests/front/Administration/CurrentMeeting.test.ts` — what an interval opening now belongs to, and the late-close guard. Full front suite green locally (343 tests).


ydiarra:
Remplacée par #6559 : la base de cette PR ne pouvait plus être changée (garde `part of a stack`, refusé par l'UI comme par les API GraphQL et REST, y compris sur le sommet de la pile). #6559 porte exactement les mêmes deux commits, rebasés sur `master` maintenant que #6549 est mergée.

PR #6553, opened 2026-09-14
feat(analytics): let the back own the meeting it already runs

ydiarra:
Base: `master`. #6549 and #6559 are merged; #6554 (in-meeting actions, Teams / Google Meet areas) is stacked on top of this one.

This is lot 0 + lot 1 of the meeting-analytics plan, extended to broadcasts.

## The problem

A meeting was reported by each participant's client opening its own interval, so a meeting of four produced four rows. Consequences, all of them live today:

- `meeting_seconds` is participant-seconds, not meeting-seconds;
- the meeting count is a `uniqExact` over an id the clients happen to agree on;
- the average number of participants is not measurable at all — a duplicated tab inflates it, a tab that dies without closing deflates it;
- `meetingProvider` is read as "what kind of meeting was this", when it only names a transport: a meeting area with ≤ `MAX_USERS_FOR_WEBRTC` participants never leaves WebRTC and was therefore filed as a spontaneous bubble;
- a broadcast was measured once per listener, by each listener.

`admin/docs/admin/generic-analytics-v3.md` already states the rule this PR follows: *avoid using participant join/leave as primary meeting counts unless the event is emitted once by a trusted authority rather than by every participant client.*

## One tracker for meetings and broadcasts

A broadcast is the same thing as a meeting: a space with people in it, open for as long as its predicate holds. Only the predicate differs — two people present, or one speaker on air. `back/src/Model/SessionAnalytics.ts` measures both:

- a session **opens when its predicate becomes true and closes when it stops holding**. A meeting area starts at the second arrival, not at the first solo visit; a broadcast from the first speaker on air to the last off. Participations are clipped to the session: a listener present before anyone went on air starts at the open;
- every row is emitted when the session **closes**, participations included — a participation cannot be emitted before we know the session it belongs to happened;
- a second session in the same space is a second row under the same id, told apart by `startedAt`. Count sessions by rows, not by `uniqExact(meetingId)`.

What the back emits:

- **`meeting.ended`** — one row per meeting, `meetingKind` (`bubble` / `area`), `participantCount`, `peakParticipantCount`. Attributed to **nobody**: naming one participant as its owner would count the whole conversation as that person's.
- **`meeting.participation.ended`** — one row per person per meeting, with `joinRank` and their own interval.
- **`broadcast.ended`** — one row per broadcast, `broadcastKind` (`megaphone` / `speaker_zone`), `participantCount`, `peakParticipantCount`, `speakerCount`. Attributed to nobody.
- **`broadcast.participation.ended`** — one row per person per broadcast, `role` (`speaker` if they were on air at any point, else `listener`), `airtimeSeconds` summed over their stints, `joinRank`.

`megaphone.ended`, `megaphone.started` and `broadcast.audience.ended` are gone from the catalog: nothing opens them any more.

## What a space is, its client says

The kind of a space is declared once at join, under the `spaceKind` metadata key, through `SpaceRegistry.joinSpace` — `bubble`, `area`, `megaphone` or `speaker_zone`, a closed zod enum shared by the front and the back (`libs/messages/src/JsonMessages/AnalyticsEventCatalog.ts`, where `meetingKind` and `broadcastKind` are cut from it). The back rejects a value outside the enum and checks nothing beyond that: the kind is the client's own claim about its space, read by the analytics and by a label the front shows itself, so a client that lies about it spoils its own rows and nothing else. A session waits for its kind before it opens.

A space that declares nothing — the world space, a chat space, a space a script opened with `WA.spaces.joinSpace` — is nobody's meeting and nobody's broadcast. That closes a false positive: an `ALL_USERS` space with `microphoneState` opened by a script would otherwise have been filed as an area meeting.

One key with an enum rather than one boolean per kind: `isMegaphoneSpace` is gone, and the UI reads the same key.

## The bubble was counted twice

`Group` registered the bubble under `${roomId}#${id}#${ts}`, but the pusher prefixes every space name with the world, so the back's `Space` for the same bubble never matched: it passed the "area" test and opened a second meeting. Every session is now owned in one place — see below — bubbles included, and `Group` has no analytics hooks left. Bubble participation rows lose `userId`, `tabId` and `clientIp` (null, as area rows already were); every row's `world` is the one the pusher writes.

A pusher restart left area participants "present" until the back itself shut down; dropping a watcher now ends their participations.

## Who owns the session, and who counts the people

`CommunicationManager` builds the session tracker, the way it has always built
`RecordingManager`. A recording and a meeting are the same shape of thing: both begin
and end inside a space, and neither is the space's own business to hold.
`ICommunicationSpace` gains `world`, `getMetadataValue` and `filterType` — what a row
needs and the transport does not.

Presence is read off the two registries the manager already keeps rather than announced
by `Space`: `users` holds who crossed the filter, `usersToNotify` holds who is watching,
and someone is in the space when they are in either. Sampling that union on both sides of
each registry mutation turns a change into the arrival or departure it is — or into
nothing, which is the common case: a speaker going on air enters `users` while already
watching, and that is not an arrival. No third roster; the two registries remember.

That needed one fix. `Space.removeUser` cleaned its own copy of `usersToNotify` and left
the manager's, so the watching half never emptied on a leave. Harmless for the transport,
which self-suppresses on a stale entry, and fatal for a duration. The normal path was
already covered — the pusher's `doUnregisterUser` sends `deleteUserFromNotify` before
`removeSpaceUserQuery` — so this only bites where the pusher never got to speak. Found
independently in #6586, whose tests came with it.

It rests on one assumption: everyone in a space watches it. True of
`ProximityChatRoom.joinSpace` and `BroadcastService.joinSpace`, both unconditional,
enforced by neither. So the manager compares `getAllUsers()` to the union and reports the
gap to Sentry, once per space, 30 s after the fact — a broadcast listener is in the space
a round trip before their client watches it, and checking on the spot would cry wolf on
every broadcast.

`Space` keeps three hooks: the on-air read, one line that reads `megaphoneState` and is
therefore indifferent to the filter where a filter crossing is not
(`LIVE_STREAMING_USERS_WITH_FEEDBACK`); the kind arriving in metadata; and closing a
session for a shutdown.

## Transport

A direct POST to the admin's existing `/api/analytics/events-batch`, not a relay through a pusher: a room and a space are watched by *every* pusher holding a client in them, so a relayed row would be emitted once per pusher — and ClickHouse would not save us, since its merges are asynchronous and no read uses `FINAL`.

The queue — bounded, batched, retried, drained on shutdown — lives once in `@workadventure/shared-utils` and is shared by the pusher and the back; the pusher keeps only what it alone does (building a row from a socket's context and checking it). `ShutdownDrains` is shared the same way.

`source: "pusher"` means "a trusted server", which is what the admin gates on: the value a socket may never claim.

## Lot 0 comes with it

The back had **no** `SIGTERM`, `SIGINT` or `uncaughtException` handling. A session only exists once it has ended, so a deploy killed the process with every live conversation inside it, silently. The gap left is SIGKILL/OOM, where the map dies with the process; persisting it would not help, since on recovery we would know a session was open but not when it ended.

## Client side

`BroadcastAnalytics.ts` is deleted. The client no longer reports meetings or broadcasts; it keeps the **list** of meetings it sends its audio to (`CurrentMeeting.ts`), because a tab can be in several at once — two areas drawn over each other, a Jitsi zone over a LiveKit area, the moment between entering an area and the bubble dissolving — and one current value had nowhere to put the second.

- A microphone or speaking period, a screen share, an action on the tab as a whole is reported **once per meeting that hears it**, cut at every meeting boundary. Exact for each meeting; a total across meetings counts an overlap twice.
- An action on one participant — mute, kick, pin, report — names that participant's space alone.
- A space the tab only receives from, a listener zone, is a broadcast and never registers: a bubble formed inside one reports to the bubble alone.

**Jitsi keeps reporting itself**: its areas join no space server-side, so the back cannot see those meetings. Those rows keep the old shape and carry no `meetingKind`, which is how the two generations are told apart in ClickHouse.

## Series impact — read before merging

`meeting_seconds` drops from participant-seconds to meeting-seconds, stacked on #6549's drop. The `startedAt` of an area meeting moves to the second arrival. Thresholds calibrated on the inflated figure need recalibrating, and the bubble/area split should move from `meetingProvider` to `meetingKind`.

**Admin MR to ship first** (SaaS repository), otherwise the new rows are dropped at ingestion and eight indicators read zero:

- `PUSHER_ONLY_EVENT_NAMES` += `broadcast.ended`, `broadcast.participation.ended`; `'broadcast.'` in `EVENT_CATEGORIES`; `broadcastId`, `broadcastKind`, `meetingKind`, `role`, `airtimeSeconds` in the anonymization-safe keys;
- readers of `megaphone.ended` / `megaphone.started` / `broadcast.audience.ended` (`megaphone_seconds`, `megaphone_started_*`, `worlds_with_broadcast`, `broadcast_airtime_seconds`, `broadcast_zone_airtime_seconds`, `broadcast_audience_seconds`, `broadcast_reach`) move to `broadcast.ended` and `broadcast.participation.ended`;
- the seeder.

## Tests

- `back/tests/SessionAnalytics.test.ts` — a bubble of three; a participation never ending after its meeting; a solo area emitting nothing; an area opening on the second arrival, closing below two and reopening as a new session; a broadcast opening on the first speaker with the audience clipped to it; airtime summed across stints; a session waiting for its kind; a session closing once and staying shut.
- `back/tests/CommunicationManager.test.ts` — a meeting measured from the kind read off the space, and nothing for a space that declared none; a member counted once whichever registry they arrive through, and kept until they have left both; the Sentry guard firing on a user in neither registry, and staying quiet when the straggler catches up before the check.
- `back/tests/Space.test.ts` — the manager is told when `removeUser` is the one dropping the `usersToNotify` entry, and is not when the pusher already unwatched.
- `back/tests/MetadataProcessor.test.ts` — a kind outside the enum is rejected.
- `play/tests/front/Administration/CurrentMeeting.test.ts` — one interval per live meeting, cut at every boundary; participant actions attributed to a conversation space and to no broadcast.
- `play/tests/pusher/AnalyticsEventCatalog.test.ts` — the back-emitted names are declared there, and the client-openable set shrinks to seven.

Full back and play suites green.


back/src/Model/Space.ts
```diff
@@ -33,6 +34,24 @@ const debug = Debug("space");
...
+ * - the megaphone space and the speaker/listener zones are broadcasts, and they are
+ *   the ones that made "connected" mean "in a meeting". They are told apart by their
+ *   filter, which is set at join time on every path.
+ * - the world space and the chat spaces carry no media to sync, so nobody is meeting
+ *   in them; this is the same predicate the communication manager already uses to
+ *   decide a space has media at all.
+ * - a proximity bubble also matches both tests above — it IS a meeting, but one the
+ *   `Group` already owns, and counting it twice is worse than not counting it here.
+ *   Its meeting is registered before the space exists, so the id is already taken.
+ */
```
moufmouf: Question: rather than tracking this in the `Group` class for the bubble conversations, we could also move analytics here?

back/src/Services/AnalyticsEventsQueue.ts
moufmouf, line 1: Question: since this happens also in the pusher, can we move the classes in charge of writing the analytics in the shared lib?

libs/shared-utils/src/ShutdownDrains.ts
moufmouf, line 1: Same thing. Duplicate code from the pusher, can be moved to shared libs.

moufmouf (changes requested):

back/src/Model/Group.ts
```diff
@@ -46,6 +46,7 @@ export class Group implements Movable, CustomJsonReplacerInterface {
         Group.nextId++;
 
         // TODO: SECURE SPACES WITH JWT tokens.
+        // The shape is what tells the back's Space it is a bubble — see isBubbleSpaceName.
```
ydiarra: inutile

libs/shared-utils/src/SpaceKind.ts
```diff
@@ -0,0 +1,20 @@
...
+ * nobody's meeting and nobody's broadcast.
+ *
+ * One key with a closed set of values rather than one boolean per kind: the front
+ * validates what it receives against this schema and the back rejects anything else,
+ * and anything inconsistent with the space's filter. Lives here rather than in
+ * `@workadventure/messages` because every file under JsonMessages feeds the API
+ * version hash, and a new value would force every connected front to reload.
+ */
+export const spaceKindSchema = z.enum(["bubble", "area", "megaphone", "speaker_zone"]);
+
```
moufmouf, line 16: This really belongs to @workadventure/messages . We don't care that it changes the version hash. Other things are already changing it anyway.

back/src/Model/MetadataProcessorInit.ts
```diff
@@ -8,6 +10,18 @@ metadataProcessor.registerMetadataProcessor("recording", () => {
...
 
+// A value outside the enum, or one that contradicts the filter the space was created
+// with, throws — and a rejected key is dropped rather than published. A meeting is a
+// space where everyone sees everyone; a broadcast is a space where only the speakers
+// are seen. A client cannot call one the other.
+metadataProcessor.registerMetadataProcessor("spaceKind", (value, _senderId, space) => {
+    const kind = spaceKindSchema.parse(value);
+    if (isMeetingKind(kind) !== (space.filterType === FilterType.ALL_USERS)) {
+        throw new Error(`A ${kind} cannot live in a space with filter ${FilterType[space.filterType]}`);
+    }
```
moufmouf, line 21: This is probably going too far. I mean... it's only for analytics. If someone manages to put an incompatible value between spaceKind and filterType.... the analytics will be wrong, that's not a huge deal. Let's remove this.

libs/shared-utils/src/SpaceKind.ts
```diff
@@ -0,0 +1,20 @@
...
+ * key: a proximity bubble, a meeting area, the world megaphone, a speaker zone. A space
+ * that declares nothing — the world space, a chat space, a space a script opened — is
+ * nobody's meeting and nobody's broadcast.
+ *
+ * One key with a closed set of values rather than one boolean per kind: the front
+ * validates what it receives against this schema and the back rejects anything else,
+ * and anything inconsistent with the space's filter. Lives here rather than in
+ * `@workadventure/messages` because every file under JsonMessages feeds the API
+ * version hash, and a new value would force every connected front to reload.
+ */
```
moufmouf, line 14: Important in the docblock: say it is used for analytics and should probably not be used for anything else (because it can be modified by clients)

back/src/Services/SpaceSessionAnalytics.ts
```diff
@@ -0,0 +1,338 @@
...
+        properties: Record<string, unknown>;
+    }): AnalyticsStoredEvent {
+        const at = new Date(input.atMs).toISOString();
+
+        return {
+            eventName: input.eventName,
+            // "pusher" means "a trusted server", which is what the admin gates on: the
+            // value a socket may never claim. A dedicated "back" would say more, but it
+            // is a JsonMessages edit, and that bumps apiVersionHash — a forced reload of
+            // every connected front, to relabel rows nothing reads by source.
```
moufmouf, line 319:
```suggestion
            // "pusher" means "a trusted server", which is what the admin gates on: the
            // value a socket may never claim.
```

back/src/Services/SpaceSessionAnalytics.ts
moufmouf, line 1: It seems me complicated compared to what it should be. I understand that it's hard to track in the `Space` if a space is used to stream or not, and the case of the megaphone says it all: we need to wait for the first user to stream to mark the Space as "onAir".

BUT! Don't we do that already in the CommunicationManager?
Aren't people streaming supposed to go through "CommunicationManager.handleUserAdded"? Could we hook on this instead (I know it's a bit more complex because we have the attendees that have video feedback here). But still... couldn't we simplify the code by putting a part of the analytics on the CommunicationManager?

PR #6554, opened 2026-09-14
feat(analytics): meeting actions, and the Teams / Google Meet areas

ydiarra:
Stacked on #6553 (itself on #6550 → #6549). Two commits: lot 3's client half, then lot 1 bis.

## 1. In-meeting actions could not be placed on a meeting

The eleven in-meeting actions — muting a microphone, muting everyone's, kicking someone, pinning them, reporting them, toggling a share, changing the layout — were declared as `signal(...)`, i.e. carrying **no properties at all**. A row said that a microphone was muted, somewhere, by someone. Same for the screen-share interval, which had its own duration and nothing to attach it to.

They all get `meetingId`, attached **centrally** in `AnalyticsClient` rather than at each call site: the answer is the same for all of them — the meeting this tab is in, which `CurrentMeeting` already tracks — and a field that has to be remembered eleven times is a field that will be forgotten once. Explicit properties still win, so the Jitsi lifecycle events remain the authority on themselves.

With this, a meeting can be read as a whole: who was in it, who had a microphone open and when, who spoke, who muted whom, who shared a screen. The admin-side endpoint that does the reading is a separate MR on the SaaS repository.

## 2. Teams and Google Meet areas were not measured at all

The Google module never called `analyticsClient`; the Teams one only for its action-bar button. The area dispatch did nothing but `inJitsiStore.set(true)` — so those visits landed in the series **as Jitsi**.

They are now reported by the client, like Jitsi and for the same reason: the back cannot see them. Their rendezvous space syncs no media, and the meeting itself happens in another tab or in the desktop app.

**So what is measured is time spent in the AREA, not in the call.** Nobody observes that tab — the module says so itself, in a `// TODO` about not being able to tell whether a meeting is in use. A user who walks away with Teams still open leaves our numbers; one who stands in the area without ever joining enters them. `meetingKind: "external"` is what keeps these rows from being read beside real meeting time without noticing, and the catalog entry spells out the rest. If you take one thing from this PR into the dashboards, take that sentence.

Gated on the module declaring its area a meeting — the same `isMeeting` flag the editor and the pathfinder already read — so a todo-list panel is not a meeting here either. A module declaring one under a subtype we have no provider for reports nothing: an unattributable row in the meeting series is worse than none, and `area.dwell` still measures the visit.

## Deliberately not fixed

Those areas still report `AvailabilityStatus.JITSI`, so `status.dwell` still calls them Jitsi. Fixing it means a new availability status: a proto change *and* a behaviour change, since the status also drives the local camera and microphone shutdown. The meeting rows carry the real provider, so the two series are separable without it — but the status series stays mixed until someone does that work deliberately.

## Tests

`play/tests/front/ExternalModule/ExternalMeetingAnalytics.test.ts` (a Teams area reported as external and cleared on leave; a module with no provider reporting nothing; a missed leave not leaving the previous meeting open) and a case in `AnalyticsClient.test.ts` for the central attachment. Front suite green: 305 passed.


libs/messages/src/JsonMessages/AnalyticsEventCatalog.ts
```diff
@@ -202,7 +202,7 @@ const meetingContextProperties = z.object({
     .describe("Identifier of the meeting, when the provider exposes one."),
   roomId: z.string().optional().describe("Room the meeting belongs to."),
   meetingProvider: z
-    .enum(["livekit", "jitsi", "webrtc"])
+    .enum(["livekit", "jitsi", "webrtc", "teams", "google_meet"])
```
moufmouf: "teams" and "google_meet" belong to an external module.
The schema here should not rely on that.

At best we could have:

```
meetingProvider: z.string()
```

or

```
meetingProvider: z.enum(["livekit", "jitsi", "webrtc","external"],
externalMeetingProviderName: z.string
```

play/src/front/ExternalModule/ExternalMeetingAnalytics.ts
```diff
@@ -0,0 +1,71 @@
...
+ * So what is measured here is **time spent in the area**, not time spent in the call.
+ * Nobody observes that tab: a user who walks away with Teams still open leaves our
+ * numbers, and one who stands in the area without ever joining enters them. The event
+ * says so, and `meetingKind: "external"` is what keeps these rows from being read
+ * beside real meeting time without noticing.
+ */
+const PROVIDERS: Record<string, "teams" | "google_meet"> = {
+    teams: "teams",
+    google: "google_meet",
+};
```
moufmouf: Same thing. We should not have "teams" or "google_meet" specific code in here. This belongs to the extensions.

PR #6562, opened 2026-09-16
Serve the Tasks Vision wasm and model as content-hashed vite assets

moufmouf:
Item 2 of the virtual background plan, stacked on #6561.

The committed copy of the Tasks Vision wasm under `play/public/static/tasksVision` (19 MB) was not the bytes shipped by `@mediapipe/tasks-vision@0.10.35`. The wasm, its loader and the segmentation model are now imported with vite's `?url`, so they are emitted under `/assets` with a content hash straight from `node_modules` and nothing is committed any more. A package bump changes the file name instead of serving a stale copy from the Cloudflare edge.

`/assets` already carried `max-age=1y`; it is now also `immutable`, so the edge never revalidates it.

Verified with a scratch vite build: `vision_wasm_internal-*.wasm` (11 153 617 bytes), `vision_wasm_nosimd_internal-*.wasm` (10 481 398 bytes), both loaders and `selfie_segmenter-*.tflite` land in `dist/public/assets` with sizes identical to the npm package.


play/src/front/WebRtc/BackgroundProcessor/tasksVisionAssets.ts
```diff
@@ -0,0 +1,22 @@
+import { FilesetResolver, type ImageSegmenter } from "@mediapipe/tasks-vision";
+// The wasm, its loader and the model are imported as vite assets: they are emitted under /assets with a content
+// hash, so a @mediapipe/tasks-vision bump can never be served a stale copy from the CDN edge, and nothing is
+// committed to the repository.
+import simdLoaderUrl from "@mediapipe/tasks-vision/vision_wasm_internal.js?url";
+import simdBinaryUrl from "@mediapipe/tasks-vision/vision_wasm_internal.wasm?url";
+import nosimdLoaderUrl from "@mediapipe/tasks-vision/vision_wasm_nosimd_internal.js?url";
+import nosimdBinaryUrl from "@mediapipe/tasks-vision/vision_wasm_nosimd_internal.wasm?url";
+import selfieSegmenterModelUrl from "./models/selfie_segmenter.tflite?url";
```
moufmouf, line 9: Question: where does this selfie_segmenter.tflite come from and can we load it directly from a node_modules package?

PR #6564, opened 2026-09-16
Segment every 2nd frame and adapt the interval to segmentation time

moufmouf:
Item 4 of the virtual background plan, stacked on #6563.

The worker clones the confidence mask and reuses it for the frames between two segmentations (default: every 2nd frame). `ResegmentController` samples segmentation time and, every 15 s, moves the interval up when the mean exceeds 30 ms or down when it is under 5 ms (bounded to 1..4), acting only when the same recommendation shows up at two consecutive checkpoints. No store, no UI.

Vitest covers the hysteresis and the mask reuse.


play/src/front/WebRtc/BackgroundProcessor/ResegmentController.ts
moufmouf: Question: we have no view on the "interval" being used. Could we maybe add a "debug" log here each time the interval is changing to check things are behaving as expected?

PR #6570, opened 2026-09-16
fix(webrtc): make the P2P reconnection actually reconnect

ydiarra:
## Context

Users report daily partial disconnects in P2P bubbles (3–4 participants): some pairs stop hearing each other after an arbitrary duration, a browser reload fixes it while a bubble rejoin is less reliable, and users behind TURN are affected more often. LiveKit meetings are stable.

A code audit of the P2P path (`SimplePeer`, `RemotePeer`, `IceServersManager`, `WebRTCCommunicationStrategy`, the `@workadventure/simple-peer` fork) found that the retry mechanism (30 attempts, backoff, new `connectionId` from the back) is only ever triggered by the *destruction* of a peer, while the two most common P2P failure modes never destroy the peer. Combined with a TURN credential lifecycle bug, this explains the reported symptoms: failures are per pair (hence partial), the duration depends on credential age (hence arbitrary), only relayed pairs die (hence TURN), and a reload fetches fresh credentials while a rejoin reuses the stale cache.

No change to the `simple-peer` fork is needed: it already exposes `iceStateChange`, `_pc`, `initiator` and `negotiate()`.

### 1. TURN credentials (front + pusher)

- **`IceServersManager.renewNow()` never re-queried the server.** It returned the cached promise as soon as one existed, so the renewal timer was a no-op: credentials were fetched once per session and expired after 4 hours. In-flight renewals are now tracked separately from the cache, and the previous set keeps being served until the new one arrives.
- `WebRTCCredentialsService`: validity raised from 4 h to 24 h. Coturn (`use-auth-secret`) checks the timestamp on every request, including the allocation *Refresh* browsers send every few minutes: once the credentials expire, a relayed connection that is still in use loses its allocation and drops. The validity must outlive the longest session, not just the connection setup.
- `TURN_CREDENTIALS_RENEWAL_TIME` default lowered from 3 h to 1 h (`.env.template`, self-hosting docs updated).

### 2. Retry only reacted to destroyed peers (front + back)

- `RemotePeer`: a peer that has not reached `connect` within 20 s is destroyed, which hands it over to the existing retry flow. Previously the connect timeouts (`RemotePeer`, `VideoBox`) only set the status to `error`, and the "persistent connection issue" message promised a reconnection that never happened.
- `SimplePeer.attemptRetry`: the restart request was one-shot. If the back never answered with a `webRtcStartMessage` (request lost, connection unknown to the back, LiveKit state...), the backoff cycle silently died. A 10 s answer timeout now schedules the next backoff attempt, without `connectionId` so that the back does not discard it as stale.
- `WebRTCCommunicationStrategy.handleMeetingConnectionRestartMessage`: when no connection is tracked but both users are in the space, a fresh connection is established instead of ignoring the request.

### 3. ICE `disconnected` was never handled (front)

The fork only destroys a peer on ICE `failed`/`closed`; `disconnected` was not observed anywhere, `connectionState` was never watched, and `restartIce()` was never used. A peer whose ICE never reaches `failed` (Firefox, dead TURN allocation) stayed muted forever with a "connected" tile.

`RemotePeer` now listens to `iceStateChange`: on `disconnected` after an established connection, the tile shows "reconnecting", the initiator calls `restartIce()` + `negotiate()` after 3 s, and the peer is destroyed (→ retry) after 15 s without recovery. `connectionstatechange === "failed"` (DTLS/SCTP) also destroys the peer. The dead `iceTimeout` listener (an event the fork never emits) is removed.

### 4. A departed user did not close the peer (front)

The back deliberately skips `webRtcDisconnectMessage` for a user who left, and `VideoBox.destroy()` keeps a peer that still receives a stream, so the `RemotePeer` lingered until ICE gave up (or forever, see 3), holding the camera and encoder. `SimplePeer` now subscribes to `observeUserLeft` and closes the peer intentionally.

### 5. Async window when replacing a peer (front)

`createPeerConnection` awaited the previous peer's promise before registering the replacement. Signals of the new `connectionId` arriving in that window hit the old peer, were rejected as a mismatch and dropped, losing the offer. The `connectionId` is now stored in the map entry so the decision is synchronous; a start superseded during its own startup returns `null` instead of reporting an error to Sentry.

### 6. `stream_stopped` triggered a spurious retry (front)

Receiving `stream_stopped` with no local stream closed the peer as a *failure*, sending a restart request and inflating `media.connection_retry`. It is now an intentional close.

## Tests

- `IceServersManager.test.ts` (new): renewal queries the server again; previous credentials are served while a renewal is in flight.
- `SimplePeer.test.ts`: next backoff attempt after an unanswered restart; departed user closes the peer without retry; replacement registered while the superseded peer is still starting up.
- `WebRTCCommunicationStrategy.test.ts`: fresh connection when tracking was lost but both users are present.

## Not covered by unit tests

The real behaviour of `restartIce()` on Chrome and Firefox needs a manual check with a simulated network cut or a killed coturn allocation.

## Operator note

For a self-hosted instance that wants every bubble on LiveKit regardless of size, `MAX_USERS_FOR_WEBRTC=0` on the back already does it (the validator accepts `0`). This PR does not add a per-world setting.


back/src/Model/Strategies/WebRTCCommunicationStrategy.ts
```diff
@@ -299,7 +299,19 @@ export class WebRTCCommunicationStrategy implements ICommunicationStrategy {
...
+            // The tracking was lost (partial cleanup) but the front still expects a connection between
+            // these two users: silently ignoring would leave them without media until one reloads.
+            const sender = this.users.get(senderUserId) ?? this.usersToNotify.get(senderUserId);
+            const receiver = this.users.get(receiverId) ?? this.usersToNotify.get(receiverId);
+            if (!sender || !receiver) {
+                console.warn(
+                    "No existing connection found for meetingConnectionRestartMessage ",
+                    senderUserId,
+                    receiverId,
+                );
```
moufmouf, line 311: Shall we Sentry this?

PR #6574, opened 2026-09-16
fix(chat): offer the reconnect prompt when the browser holds no Matrix session

ydiarra:
## Summary

- A user logged in to WorkAdventure whose browser has lost its Matrix credentials (storage evicted, session revoked on an earlier visit) got a bare "Chat not available" banner on every page load, with nothing to click. Only a new OpenID login can mint the Matrix login token, and the front never offered it.
- Visible in Sentry as [PLAY-FG](https://workadventure-e6.sentry.io/issues/PLAY-FG) ("Unable to connect to matrix, access token is null", ~100 events a day). Reported by a cube.fr user stuck since 2 September on Safari.
- The missing-session state is now typed as `MissingMatrixCredentialsError` and handled like an invalid login token: the chat panel shows the "reconnect" prompt, whose button goes through the OpenID flow.

## Test plan

- [x] `vitest run src/front/Chat/Connection/Matrix/tests/MatrixClientWrapper.test.ts` (8 passed, the null-token test now also asserts the error type)
- [x] eslint + prettier on the changed files
- [ ] Manual: log in, clear the `matrix*` keys from localStorage, reload: the chat panel shows the reconnect prompt instead of only the red banner


moufmouf (approved):
LGTM

PR #6576, opened 2026-09-17
fix(webrtc): let Firefox pick the resolution itself, its VP9 encoder dies on odd frame sizes

ydiarra:
## Symptom

Entering a bubble from Firefox, the peers receive **one frame and nothing more**. The sender's own stats say it all: resolution `0x0`, 0 fps, frames-encoded counter frozen for good, while `setParameters` reports success and the connection stays up. The viewer meanwhile shows "Expected FPS 20 / FPS 0", so it looks like a stalled link when it is really a dead encoder.

## Cause

**Firefox's VP9 encoder stops producing frames permanently as soon as a scaled frame has an odd width or height.** VP9 is the first codec we negotiate for cameras, and the adaptive encoding scales the sent video down to the viewer's tile with an arbitrary `scaleResolutionDownBy` ratio, which lands on an odd size for most tile sizes.

Crucially, the scale applies to the frames **the encoder really receives**, whose size we do not control: a camera ramps its resolution up and down on its own (Firefox starts at 960x540 and climbs to 1280x720, and drops back down under a bitrate cap). So no scale computed from `track.getSettings()` is safe for long. A scale of 7.33 gives an even 174x98 from 1280x720 but **130x73** from 960x540.

Measured, Firefox 157 to Chromium over P2P with a real camera:

| Sender settings | Result |
|---|---|
| scale even for 1280x720, source really 960x540 | encoder dead, `0x0`, counter frozen |
| no scale, bitrate cap only, source 960x540 | streams, own ladder 320x180 → 480x270 → 640x360 |
| no scale, bitrate cap only, source 1280x720 | streams, drops 960x540 → 640x360 when capped |
| same sequence in VP8 | fine |

## Fix

- **Drop `scaleResolutionDownBy` on Firefox**, keep the bitrate and frame rate caps. Its own quality scaler already picks even sizes and honours `maxBitrate`, which is what the scale was there for. Chrome keeps scaling exactly as before.
- Start the "cut the video unless the viewer reports its tile" countdown **once the connection is up**. Firefox has `encodings` before negotiation (Chrome does not), so the 5 second timer used to start when the track was added, and a slow connection paused the encoder before the viewer had any chance to report.

The first two commits of this branch tried to snap the scale to even sizes; the third replaces that with the above, since snapping cannot hold when the source resolution moves.

## Tests

- `AdaptiveVideoEncoding.test.ts` covers the new flag: no scale when it is off, bitrate and frame rate caps still applied, encoder still stopped for a hidden viewer.
- `adaptive_streaming.spec.ts` is still `@nofirefox`, which is why this whole path was never covered on Firefox.


moufmouf:
Superseded by #6582

PR #6578, opened 2026-09-17
feat(oidc): renew the provider's access token with a refresh token

dfr-exnaton:
## Problem

WorkAdventure keeps only the access token from the OpenID Connect provider and re-validates it against the userinfo endpoint on every page load. Access tokens are typically short-lived (one hour on Google) while WorkAdventure's own auth token lasts 30 days, so the session silently dies at the provider's expiry: the user is still playing, but the next reload bounces them back to the login screen.

## Fix

Ask the provider for a refresh token and use it to renew the access token instead of forcing a re-login:

- `OPENID_ACCESS_TYPE` lets a deployment send `access_type=offline` on the authorization request. Google needs this Google-style parameter and ignores the `offline_access` scope that spec-compliant providers expect in `OPENID_SCOPE`. Left unset, the parameter is not sent and behavior is unchanged.
- The refresh token is encrypted with `SECRET_KEY` before going into the auth token, which lives in the browser's `localStorage` with a plain base64 payload.
- `checkTokenAuthWithRefresh` renews the access token when, and only when, the provider rejects the current one (`invalid_token` / `invalid_grant` / 401), so network and configuration failures still surface instead of being retried away.
- `/me` answers with a re-minted auth token when the access token was renewed, which the front already persists via `localUserStore.setAuthToken`.
- Rotated refresh tokens are stored; providers that do not rotate keep the current one.
- Logout also revokes the refresh token, since revoking an access token does not revoke it on a spec-compliant provider.

## Testing

- `npm run typecheck`, `npm run lint`, `npm run pretty-check` (play)
- `npm test -- --run tests/pusher` (play) — includes 7 new unit tests covering the refresh paths
- `npm run check-env-docs`


moufmouf:
Oh excellent! Encrypting the refresh token is the part that has been missing from previous attempts like this one: #6098 

We just need a way to test this easily. Could you update the PR so that the test OIDC server shipped with WorkAdventure has an access token with a much shorter life (like 2-3 minutes) so we can test the refresh token actually works?

dfr-exnaton:
Thanks for your quick response and the awesome tool! I actually didn't see the previous attempt 🙈  - as the code is mostly LLM written, I will review/compare with your attempt and add testing.

I can confirm that a manual patch of this works, at least with Google...

PR #6586, opened 2026-09-21
fix(space): clear the communication registry when removeUser drops a user to notify

ydiarra:
Stacked on #6553.

## The leak

`Space.removeUser` deletes the leaving user from its watcher's `usersToNotify` map, but its `finally` only calls `communicationManager.handleUserDeleted(user)` (and only when the user passes the filter). It never calls `handleUserToNotifyDeleted`, so the `UserRegistry` inside `CommunicationManager` keeps the departed user in its *users to notify* collection.

The entry is then unrecoverable: `removeWatcher` rebuilds its notify teardown from `this.usersToNotify.get(watcher)`, and `removeUser` has already emptied it. The registry stays stale for the lifetime of the space.

## Why it matters

* `WebRTCCommunicationStrategy.deleteUser` skips `shutdownAllConnections(user)` when `usersToNotify.has(user)` — the leaver still looks like a watcher, so its peer connections are never torn down, and `cleanupUserMessages` only drops the forward direction.
* Any later `initialize()` walks `users × usersToNotify` and calls `establishConnection(B, ghost)`: the remaining users receive a `webRtcStartMessage` for someone who left. LiveKit's `initialize()` likewise runs `addUserToNotify(ghost)` and re-invites it.

Not affected: `TransitionPolicy` / `MAX_USERS_FOR_WEBRTC` read `space.getAllUsers().length`, never the registry (`UserRegistry.getTotalCount()` has no production caller), and `stopRecordingIfUserCompletelyLeftSpace` reads the space maps, which stay correct.

## The fix

Keep the return value of the defensive `usersToNotifyList.delete()` and, when it actually removed something, call `handleUserToNotifyDeleted` — ungated by the filter, symmetric with `addUserToNotify`, and before `handleUserDeleted` so the teardown sees the same ordering the pusher uses.

The pusher's `doUnregisterUser` sends `deleteSpaceUserToNotify` before `removeSpaceUserQuery` on the same stream, so on the normal leave path the delete is a no-op and behaviour is unchanged.

## Tests

Two cases in `back/tests/Space.test.ts`: the manager is notified when `removeUser` is the one dropping the entry, and is not notified when the pusher already unwatched. The first fails without the fix. `npm run typecheck` and `npm test` pass in `back/`.


ydiarra:
Absorbée dans #6553.

Le correctif est arrivé des deux côtés en parallèle : il est dans `1d6aa0485` (`refactor(analytics): read presence off the registries the manager already keeps`), à l'identique — même `wasToNotify` sur le retour de `delete()`, même appel non filtré, même ordre avant `handleUserDeleted`.

Il ne pouvait pas rester empilé **sur** #6553 : cette PR y dérive désormais la présence de l'union `users` ∪ `usersToNotify`, donc elle a besoin que le registre soit nettoyé au départ — le correctif doit être dedans, pas au-dessus.

Les deux tests de `back/tests/Space.test.ts` sont repris tels quels dans #6553 (`80721bcf93`), y compris le second, qui est le vrai apport : il documente que `doUnregisterUser` envoie déjà `deleteUserFromNotify` avant `removeSpaceUserQuery`, donc que le correctif ne mord que sur les chemins où le pusher n'a pas eu le temps de parler. J'avais conclu l'inverse de mon côté.

#6587 reste ouverte et indépendante.

